PR #14, opened 2018-04-06
Code cov support

### LINKED ISSUE #13: Add CodeCov Support
One relevant thing could be integrate code coverage in the project.
Mode info: https://github.com/codecov/example-swift

What do you think?

I can do it if you are agree.

ConradoMateu:
Added Support in Travis in order to enable Code Coverage support.

close #13


ConradoMateu:
If it's public I think you don't need it, thats just for private repos :)

victorg1991:
Fair enough, thanks for the clarification :D

PR #16, opened 2018-04-16
Naming changes: Kind and fix()

truizlop:
This PR changes naming to align with the conventions in Arrow, addressing part of issue #17. The main changes that have been done are:
- `HK` has been renamed to `Kind`. Similarly, `HKn` to `Kindn`, for n = 2..5
- Witnesses ended in F have been renamed to start with `For-`; e.g. `MaybeF` is now `ForMaybe`.
- Static methods and extension methods named `ev` before are now `fix`.
- Data classes ending in `KW` now end in `K`.
- Aliases for higher kinds have been introduced; e.g. `Kind<ForMaybe, A>` is `MaybeOf<A>`.


pakoito:
>Witnesses ended in F have been renamed to end in Kind; e.g. MaybeF is now MaybeKind.

In Arrow we renamed them to start with `For-`, like `ForMaybe`. The feedback has been positive.

raulraja:
In favor of keeping Arrow names if they are ok in swift land with community expectations

truizlop:
No problem, I can change them to For* prefixes

victorg1991:
I like that naming too and I think it's ok in swift land :)

truizlop:
@pakoito @raulraja @victorg1991 just changed the prefixes and fixed CI

Sources/Bow/Data/ListKW.swift
```diff
@@ -175,117 +175,117 @@ public extension ListKW {
...
-    public func traverse<G, A, B, Appl>(_ fa: HK<ListKWF, A>, _ f: @escaping (A) -> HK<G, B>, _ applicative: Appl) -> HK<G, HK<ListKWF, B>> where G == Appl.F, Appl : Applicative {
-        return fa.ev().traverse(f, applicative)
+    public func traverse<G, A, B, Appl>(_ fa: Kind<ForListKW, A>, _ f: @escaping (A) -> Kind<G, B>, _ applicative: Appl) -> Kind<G, Kind<ForListKW, B>> where G == Appl.F, Appl : Applicative {
+        return fa.fix().traverse(f, applicative)
     }
 }
 
 public class ListKWSemigroupK : SemigroupK {
-    public typealias F = ListKWF
+    public typealias F = ForListKW
```
raulraja: All types that were previously suffixed by `KW` are now just `K` so `ListKW` becomes `ListK`

Sources/Bow/Data/ListKW.swift
```diff
@@ -175,117 +175,117 @@ public extension ListKW {
...
     
-    public func combineK<A>(_ x: HK<ListKWF, A>, _ y: HK<ListKWF, A>) -> HK<ListKWF, A> {
-        return x.ev().combineK(y.ev())
+    public func combineK<A>(_ x: Kind<ForListKW, A>, _ y: Kind<ForListKW, A>) -> Kind<ForListKW, A> {
```
raulraja: Assuming type aliases work in the same way as Kotlin we can replace in the codebase all instances of types such as `Kind<ForListKW, A>` for their abbreviated aliased form `ListKOf<A>` which is much more readable.

truizlop:
@raulraja I addressed your comments in the last list of commits. The code is much more readable now with those aliases

raulraja:
We should consider at this point once this is merged if it makes sense to do codegen to ease the generation of kinds and instances. Also in Arrow 0.7.0 the encoding of type classes uses extensions for methods whose first args used to be a kind to enable direct syntax in blocks where the type classes are in scope. Not sure if that is possible or desirable in swift but if it is then it's another area where we want to go in line with Arrow's encoding

PR #19, opened 2018-04-19
Remove headers

truizlop:
This PR removes header comments created by Xcode by default.


victorg1991:
@truizlop we have to add another header, similar to the arrow one, don't we?

truizlop:
@victorg1991 I've checked several files in the Kotlin repo and they don't have the license comments on top, only on the README. I suggest we proceed accordingly. I'll add the license to the README as well, as part of this PR.

victorg1991:
Perfect then :)

Issue #23, opened 2018-04-30
Optics


truizlop:
Missing things regarding optics:

- Typeclasses: `At`, `Each`, `FilterIndex`, `Index`
- Optics: `Fold`, `Getter`, `Iso`, `Lens`, `Optional`, `Prism`, `Setter`, `Traversal`
- Laws for all optics

RPallas92:
Reference implementation for [Lens](https://github.com/facile-it/FunctionalKit/blob/master/Sources/FunctionalKit/Lens.swift) and for [Prism](https://github.com/facile-it/FunctionalKit/blob/master/Sources/FunctionalKit/Prism.swift)

truizlop:
#24 implemented `Getter`, `Iso`, `Lens`, `Optional`, `Prism` and `Setter`. Also, all their corresponding laws were included in that PR. Still need to test all their methods though.

truizlop:
#25 implemented `Fold` and `Traversal`, with their corresponding laws.

truizlop:
#26 extends data types with optics.
#27 provides instances for optics.
#28 extends optics typeclasses.
#30 adds tests to optics.

PR #24, opened 2018-05-04
Add initial implementation for Optics

truizlop:
This PR implements some of the Optics present in Arrow. The following optics are fully implemented:
- Iso
- Lens
- Getter
- Setter
- Prism
- Optional

An initial implementation for Fold is included, but not completed. Also, Traversal is not included.

For the implemented optics, their corresponding laws have been encoded in the test folder, and those laws were checked for each one of them with a simple implementation.

It needs more testing, but I didn't want to make the PR much bigger than this.


Sources/Bow/Optics/Fold.swift
```diff
@@ -0,0 +1,72 @@
+import Foundation
+
+public class ForFold {}
+public typealias FoldOf<S, A> = Kind2<ForFold, S, A>
+public typealias FoldPartial<S> = Kind<ForFold, S>
+
+public class Fold<S, A> : FoldOf<S, A> {
```
nomisRev: Missing all composition functions.

truizlop:
As stated in the text of the PR, I faced some issues in Swift implementing Fold and Traversal, that's why they are not part of this PR. If you prefer, I can remove the partial implementation of Fold and add it later once it is complete. What do you think @nomisRev?

nomisRev:
@truizlop Sorry for the late response, this completely slipped my mind.

Yes, feel free to merge already. Perhaps update #23 with status.

I faced some issues with `Fold` and `Traversal` when I was implementing them. I don't have a lot of experience with Swift but if you want we can discuss the issues further on Gitter.

truizlop:
Thanks @nomisRev, we can discuss on gitter the details for `Fold` and `Traversal`. I need to find time to implement them, but as soon as I get stuck, I'll contact you.

PR #25, opened 2018-06-21
Missing optics

truizlop:
Pull request #24 addressed an initial implementation for some of the optics present in Arrow. This PR completes the list with: 

- Implementation of Fold and Traversal.
- Conversions and compositions between each optics. 
- Laws for Traversal and tests for these laws.

There are still some tests missing for each optic to make sure they work properly.

### DIFF
--- a/Sources/Bow/Optics/Fold.swift
+++ b/Sources/Bow/Optics/Fold.swift
@@ -58,6 +58,10 @@ open class Fold<S, A> : FoldOf<S, A> {
         return foldMap(ListK<A>.monoid(), s, ListK<A>.pure).fix()
     }
     
+    public func choice<C>(_ other : Fold<C, A>) -> Fold<Either<S, C>, A> {
+        return ChoiceFold(first: self, second: other)
+    }
+    
     public func find(_ s : S, _ predicate : @escaping (A) -> Bool) -> Maybe<A> {
         return foldMap(FirstMaybeMonoid<A>(), s, { a in predicate(a) ? Const<Maybe<A>, First>(Maybe.some(a)) : Const(Maybe.none()) }).value
     }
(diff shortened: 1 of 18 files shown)


truizlop:
I agree. @nomisRev can you take a look? I'd like to continue with other missing parts, but I need this to be merged first.

nomisRev:
@truizlop I'll review it today!

Sources/Bow/Optics/Fold.swift
```diff
@@ -4,14 +4,58 @@ public class ForFold {}
 public typealias FoldOf<S, A> = Kind2<ForFold, S, A>
 public typealias FoldPartial<S> = Kind<ForFold, S>
 
-public class Fold<S, A> : FoldOf<S, A> {
+open class Fold<S, A> : FoldOf<S, A> {
```
nomisRev: All optics can be `protocols`. I missed that in the previous review.

Sources/Bow/Optics/Fold.swift
```diff
@@ -4,14 +4,58 @@ public class ForFold {}
...
+    public static func +<C>(lhs : Fold<S, A>, rhs : Optional<A, C>) -> Fold<S, C> {
+        return lhs.compose(rhs)
+    }
+
+    public static func +<C>(lhs : Fold<S, A>, rhs : Traversal<A, C>) -> Fold<S, C> {
+        return lhs.compose(rhs)
+    }
+    
+    open func foldMap<Mono, R>(_ monoid : Mono, _ s : S, _ f : @escaping (A) -> R) -> R where Mono : Monoid, Mono.A == R {
+        fatalError("foldMap must be overriden in subclasses")
```
nomisRev: No need for `fatalError` if `Fold` is a protocol.

Sources/Bow/Optics/Traversal.swift
```diff
@@ -0,0 +1,614 @@
...
+
+public typealias Traversal<S, A> = PTraversal<S, S, A, A>
+public typealias ForTraversal = ForPTraversal
+public typealias TraversalOf<S, A> = PTraversalOf<S, S, A, A>
+public typealias TraversalPartial<S> = Kind<ForPTraversal, S>
+
+open class PTraversal<S, T, A, B> : PTraversalOf<S, T, A, B> {
+
+    open func modifyF<Appl, F>(_ applicative : Appl, _ s : S, _ f : @escaping (A) -> Kind<F, B>) -> Kind<F, T> where Appl : Applicative, Appl.F == F {
+        fatalError("modifyF must be implemented in subclasses")
```
nomisRev: Same story as `Fold`.

Sources/Bow/Optics/Fold.swift
truizlop: The problem we have is that in Swift we cannot make those protocols be higher kinds, and the only way to implement Fold and Traversal was through extension, because of the way Swift handles generics and protocols with associated types.

PR #26, opened 2018-07-09
Extend datatypes with optics

truizlop:
This PR extends some of the datatypes with optics. Some still are missing since their corresponding datatypes are not implemented.


truizlop:
@nomisRev we need your review here as well

nomisRev:
Looks good! :+1:

PR #28, opened 2018-07-10
Extend optics typeclasses

truizlop:
This PR extends the existing typeclasses for optics (namely `At`, `Each`, `Index` and `FilterIndex`) with static methods to create implementations for them, from `Iso` and `Traverse`.

### DIFF
--- a/Sources/Bow/Optics/Typeclasses/At.swift
+++ b/Sources/Bow/Optics/Typeclasses/At.swift
@@ -7,3 +7,32 @@ public protocol At : Typeclass {
     
     func at(_ i : I) -> Lens<S, A>
 }
+
+public extension At {
+    public static func at<AtType>(_ at : AtType, _ i : I) -> Lens<S, A> where AtType : At, AtType.S == S, AtType.I == I, AtType.A == A {
+        return at.at(i)
+    }
+    
+    public static func fromIso<U, AtType>(_ at : AtType, _ iso : Iso<S, U>) -> AtFromIso<S, I, A, U, AtType> where AtType : At, AtType.S == U, AtType.I == I, AtType.A == A {
+        return AtFromIso(at: at, iso: iso)
+    }
+}
+
+public class AtFromIso<M, N, O, P, AtType> : At where AtType : At, AtType.S == P, AtType.I == N, AtType.A == O {
+    public typealias S = M
+    public typealias I = N
+    public typealias A = O
+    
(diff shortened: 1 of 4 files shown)


Sources/Bow/Optics/Typeclasses/At.swift
```diff
@@ -7,3 +7,32 @@ public protocol At : Typeclass {
...
+    public static func at<AtType>(_ at : AtType, _ i : I) -> Lens<S, A> where AtType : At, AtType.S == S, AtType.I == I, AtType.A == A {
+        return at.at(i)
+    }
+    
+    public static func fromIso<U, AtType>(_ at : AtType, _ iso : Iso<S, U>) -> AtFromIso<S, I, A, U, AtType> where AtType : At, AtType.S == U, AtType.I == I, AtType.A == A {
+        return AtFromIso(at: at, iso: iso)
+    }
+}
+
+public class AtFromIso<M, N, O, P, AtType> : At where AtType : At, AtType.S == P, AtType.I == N, AtType.A == O {
```
nomisRev: I don't quite understand why this is required. Teach me please!

nomisRev: The return types from `fromIso` seem weird because of this.

truizlop: There are several issues in Swift. First one is you cannot create anonymous implementations of protocols (interfaces); therefore, you need to create an external class for that. In this case, I'd like to make something like we do in Kotlin: `object : At { ... }`, but in Swift it is not possible.

Second problem has to do with protocols with generic types. In Java/Kotlin we can have an interface `Interface<A, B>`; in Swift, that translates to:

```swift
protocol Interface {
  associatedtype A
  associatedtype B
}
```

These are known as protocols with associated types. Moreover, these protocols cannot be directly used in the signature of a function:

```swift
func whatever(x : Interface) { ... } // This does not compile
```

Instead, the language forces you to add a generic type and use the protocol with associated types as a constraint over that type:

```swift
func whatever<T>(x : T) where T : Interface { ... }
```

The problem gets even worse when these generic types are used as a return type; most of the time the compiler is not able to match the inferred types and things do not compile.

As you can see, the syntax gets pretty cumbersome, but so far it is the best we can get with how the language is designed. I hope it changes in the future, but for now we will have to live with that.

PR #30, opened 2018-07-19
Test Optics

truizlop:
This PR adds tests properties of all Optics, as well as checks the composition between each other works as expected.


truizlop:
@nomisRev I need your review here, although they are mostly tests

PR #32, opened 2018-07-20
Conformance to CustomDebugStringConvertible

truizlop:
This PR adds conformance to `CustomDebugStringConvertible` to the following data types:

- `Const`
- `Either`
- `Id`
- `Ior`
- `ListK`
- `Maybe`
- `NonEmptyList`
- `Try`
- `Validated`

Conformance to this protocol provides an easier understanding of the contents of a data type when debugging it.

### DIFF
--- a/Sources/Bow/Data/ListK.swift
+++ b/Sources/Bow/Data/ListK.swift
@@ -125,6 +125,20 @@ public extension Array {
     }
 }
 
+extension ListK : CustomStringConvertible {
+    public var description : String {
+        let contentsString = self.list.map { x in "\(x)" }.joined(separator: ", ")
+        return "ListK(\(contentsString))"
+    }
+}
+
+extension ListK : CustomDebugStringConvertible where A : CustomDebugStringConvertible {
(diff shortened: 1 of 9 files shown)


Sources/Bow/Data/ListK.swift
```diff
@@ -125,6 +125,20 @@ public extension Array {
...
+extension ListK : CustomStringConvertible {
+    public var description : String {
+        let contentsString = self.list.map { x in "\(x)" }.joined(separator: ", ")
+        return "ListK(\(contentsString))"
+    }
+}
+
+extension ListK : CustomDebugStringConvertible where A : CustomDebugStringConvertible {
+    public var debugDescription : String {
+        let contentsString = self.list.map { x in x.debugDescription }.joined(separator: ", ")
```
victorg1991: why not just self.list.debugDescription ?

truizlop: Do you know if that goes into every object debugDescription? If so, I can change it.

PR #34, opened 2018-07-23
Add SetK | Fixes https://github.com/arrow-kt/bow/issues/21

victorg1991:


truizlop:
I've been doing some attempts but I think it is not possible, unless we don't rely on Swift internal implementation for Set and make our own. It think, for now, we can leave SemigroupK, MonoidK and Foldable out (perhaps make an issue to remind ourselves this needs to be addressed) and merge it.

Everything looks ok to me; just some minor things: sort by name both production and test Data folders and make sure the SetK file is in all four targets (right now it is only on the main Bow target).

victorg1991:
Done! :D

PR #35, opened 2018-07-23
Configure travis to make a new release when pushing a tag

victorg1991:
I have to test it first, but it should release a new version of the library when pushing a new tag. 
You would have to manually update the version number in the project and in the .podspec tho.

I would like to include fastlane to handle all of this too :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: objective-c
 osx_image: xcode9.3
-
+env:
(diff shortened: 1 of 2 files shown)


pakoito:
This is too easy. Where's the maven registration process? The token and secret dance? REJECT, REJECT!

truizlop:
+1 to include fastlane to handle that. Can you make an issue and take care of that @victorg1991 ?

victorg1991:
Nice thanks! @pakoito you jealous? :trollface:

PR #39, opened 2018-07-31
Add file to all targets

### LINKED ISSUE #38: Fetching the library through Carthage fails to build project
It looks like not all the files are included in every target; therefore, building the library for other targets (like watchOS) fails to find some of the files.

truizlop:
Fixes #38. `BoolInstances` was only included in the main target and caused other targets not to compile.


truizlop:
@victorg1991 although this solves the issue, perhaps we need to modify Travis to build all targets and warn us if we forget to include some file in some of the targets.

PR #40, opened 2018-07-31
Make travis build all four targets to make sure no file is missing

truizlop:
The previous Travis configuration was building only the main target (for iOS). With the new configuration, Travis will build additionally the targets for macOS, watchOS and tvOS, making sure all files are included in all targets and preventing issues like #38.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,22 @@
 language: objective-c
-osx_image: xcode9.3
+osx_image: xcode9.4
 env:
+  global:
+    - PROJECT=Bow.xcodeproj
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,18 +1,27 @@
 language: objective-c
-osx_image: xcode9.3
+osx_image: xcode9.4
 env:
+  global:
+    - PROJECT=Bow.xcodeproj
   secure: LatJH8j809/6dhEO6WhpQyU3lAjtkOYHAqWD14+bU6lfAc5oRrRFO4B7h56gnIDu6tvYJS0Qa/IhIuB3zZQ/GlG12oivgE/FKp1xXxxHwqjmhPHdetX9YkeAbicJ8YrSQ+C0UC6NMmP1WtU7cmLBV4Q1VwdPnVvRIrOVuWdh3ITUzUz/qQMTZwi+6Z1w92QvxbyytXDqti62NrUq81ba2c7pvBbFBujMp5hyd6+EcuZojYWGNN3cIR2DcPVXD9j4kwLyDPVAcN/OROAwaJlOZ63FbsoUcZULOAX/6s5bmcU1CuYc+eBrXX7KifSqj3yUW00Ea0ynyvsn2GWldAw5b8cYgiLuqQBp3jxtPVH/tEWRdO5TdFtMl8CDEJQiblINjFCXRbnAh8fnk/9jxKTkK+e1DXNQEk/w61+vUB21j74Zw8PhPRvCsoV8XgPkd34UiSlG5XkPaj+vImRpB2/kemYxynpY2Sjq+qoK4iUJy/DXhX/agThWCpjY1hWL2LNL/24mk6DbUHZGVH+4RgoBJxBFCI7ucSp14Og0cCFmhFfsq0nttN+xNegosmHnEcv1Z4tRW750rPf2dgGlXkf6s4msbZ/kyRnccs2zlh3JYYac509Fk98FRNfLb9q0C/jLb0xpg6qf9WbI1QbMcxPHKakyGEUuo+7FO+CLzrJ0qns=
+matrix:
+  include:
+    - env: JOB="XCODE" DEST="OS=11.4,name=iPhone 8 Plus" SCHEME="Bow" SDK="iphonesimulator" ACTION="test"
```
victorg1991: This JOB env var is not used, is it needed for something?

.travis.yml
```diff
@@ -1,18 +1,27 @@
...
+    - env: JOB="XCODE" DEST="arch=x86_64" SCHEME="Bow-macOS" SDK="macosx" ACTION="test"
+    - env: JOB="XCODE" DEST="OS=11.4,name=Apple TV" SCHEME="Bow-tvOS" SDK="appletvsimulator" ACTION="test"
+    - env: JOB="XCODE" DEST="OS=4.3,name=Apple Watch - 42mm" SCHEME="Bow-watchOS" SDK="watchsimulator" ACTION="build"
+
 before_install:
   - carthage bootstrap
```
victorg1991: Now that we have 4 jobs, we can improve this and build only the needed platform in each case with the flag --platform iOS MacOS etc. 

But we can do this in a separate PR :)

.travis.yml
truizlop: You are right, I copied it from an old script and forgot to remove it

victorg1991:
Mmm, I tried locally and it was failing too, I have removed the code sign and it seems to work, let's see on travis

truizlop:
I built all 4 targets locally and it was working, but on CI it is still failing due to problems in code signing. Any idea about what we can do?

PR #46, opened 2018-10-01
Add missing core typeclasses

truizlop:
This PR adds some missing typeclasses from the core module in Arrow; in particular:

- Bifunctor
- Category
- Contravariant
- Invariant
- Profunctor

In addition, each typeclass has its own laws for testing.


Sources/Bow/Typeclasses/Category.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+public protocol Category {
```
pakoito: Do we have this in Arrow?

truizlop: I found it there, in the core module

PR #47, opened 2018-10-01
Add effects typeclasses

truizlop:
This PR adds typeclasses in the Effects module from Arrow; more precisely:

- Async
- ConcurrentEffect
- Effect
- MonadDefer


Sources/Bow/Effects/IO.swift
```diff
@@ -351,7 +351,11 @@ public class IOMonad : IOApplicative, Monad {
     }
 }
 
-public class IOAsyncContext : AsyncContext {
+public class IOAsync : IOMonadError, Async {
+    public func deferExecution<A>(_ fa: @escaping () -> Kind<ForIO, A>) -> Kind<ForIO, A> {
```
raulraja: unless it collides with existing keyword this operation is frequently named `suspend` and it's variation that automatically calls `pure` or `raiseError` is frequently called `delay`

Sources/Bow/Effects/Typeclasses/MonadDefer.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+public protocol MonadDefer : MonadError where E == Error {
+    func deferExecution<A>(_ fa : @escaping () -> Kind<F, A>) -> Kind<F, A>
+}
+
+public extension MonadDefer {
+    public func invoke<A>(_ f : @escaping () throws -> A) -> Kind<F, A> {
```
raulraja: This should be called `delay` and it will be also renamed in Arrow in 0.8.0 since the invoke operator creates a lot of ambiguity

Sources/Bow/Effects/Typeclasses/MonadDefer.swift
```diff
@@ -0,0 +1,26 @@
...
+        return self.deferExecution {
```
raulraja: `suspend`

Sources/Bow/Effects/Typeclasses/MonadDefer.swift
```diff
@@ -0,0 +1,26 @@
...
+                return self.raiseError(error)
+            }
+        }
+    }
+    
+    public func lazy() -> Kind<F, Unit> {
+        return invoke {}
+    }
+    
+    public func deferUnsafe<A>(_ f : @escaping () -> Either<Error, A>) -> Kind<F, A> {
```
raulraja: AFAIK this is not unsafe but pure because `suspend` and `delay` defer execution of any effects.

truizlop: I thought the same, but I used the same name as in Arrow. I've changed it to `delayEither` and we may think a better name later.

PR #49, opened 2018-10-03
Comonadic UIs data types

truizlop:
This PR adds several data types described in the paper [Declarative UIs are the Future - and the Future is Comonadic!](https://functorial.com/the-future-is-comonadic/main.pdf) and implemented as in [this PR in Arrow](https://github.com/arrow-kt/arrow/pull/1020). The included data types are:

- Store
- Moore
- Sum
- Day

The only missing things with respect to the Arrow version are the `lazy` versions of methods in `Day`, and the applicative instances. I had several problems with types and generics and couldn't implement them in a safe way. I'd probably need to revisit this in the future.

### DIFF
--- a/Sources/Bow/Data/Sum.swift
+++ b/Sources/Bow/Data/Sum.swift
@@ -2,6 +2,7 @@ import Foundation
 
 public class ForSum {}
 public typealias SumOf<F, G, V> = Kind3<ForSum, F, G, V>
+public typealias SumPartial<F, G> = Kind2<ForSum, F, G>
 
 public enum Side {
     case left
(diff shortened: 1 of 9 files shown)


Sources/Bow/Data/Sum.swift
```diff
@@ -0,0 +1,105 @@
...
+public class ForSum {}
+public typealias SumOf<F, G, V> = Kind3<ForSum, F, G, V>
+public typealias SumPartial<F, G> = Kind2<ForSum, F, G>
+
+public enum Side {
+    case left
+    case right
+}
+
+public class Sum<F, G, V> : SumOf<F, G, V> {
```
raulraja: Down the road, we may have to change the name of this data type here and in Arrow since it is definitely confusing when thinking about the `Sum` `Monoid<Number>` which we still need a strategy to disambiguate from the `Product` one

truizlop: I agree, it was confusing to me when I ported it from Arrow.

PR #51, opened 2018-10-05
Add validation functions

truizlop:
This PR adds multiple validation functions as in the equally named package in Arrow.

### DIFF
--- a/Sources/Bow/Data/Either.swift
+++ b/Sources/Bow/Data/Either.swift
@@ -105,6 +105,10 @@ public class Either<A, B> : EitherOf<A, B> {
     public func combineK(_ y : Either<A, B>) -> Either<A, B> {
         return fold(constant(y), Either<A, B>.right)
     }
+    
+    public func get() -> B {
+        return fold({ _ in fatalError("Attempted to get a right value but got \(self.description)") }, id)
+    }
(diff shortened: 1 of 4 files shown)


Sources/Bow/Data/Either.swift
```diff
@@ -105,6 +105,10 @@ public class Either<A, B> : EitherOf<A, B> {
     public func combineK(_ y : Either<A, B>) -> Either<A, B> {
         return fold(constant(y), Either<A, B>.right)
     }
+    
+    public func get() -> B {
```
pakoito: We're on the way of deprecating this. I'd suggest not adding it.

Sources/Bow/Validation/Validation.swift
```diff
@@ -0,0 +1,475 @@
+import Foundation
+
+public class Validation<E> {
```
pakoito: The validation framework is a remnant from Funktionale that's in it's way out for not being lawful. I'd also not suggest not adding it.

truizlop:
Ok, so probably better to remove this whole branch, right?

pakoito:
Yes, probably. I thought we had deprecation notices in place to flag it 🙈

PR #53, opened 2018-10-15
BrightFutures Effect integration

truizlop:
This PR adds an integration with [BrightFutures](https://github.com/Thomvis/BrightFutures), a popular iOS library for futures. It wraps the `Future` class into `FutureK` and provides the necessary instances to work with it.


Sources/Bow/Effects/BrightFutures/FutureK.swift
```diff
@@ -0,0 +1,203 @@
...
+    public func flatMap<A, B>(_ fa: FutureKOf<E, A>, _ f: @escaping (A) -> FutureKOf<E, B>) -> FutureKOf<E, B> {
+        return FutureK<E, A>.fix(fa).flatMap(f)
+    }
+    
+    public func tailRecM<A, B>(_ a: A, _ f: @escaping (A) -> FutureKOf<E, Either<A, B>>) -> FutureKOf<E, B> {
+        return FutureK<E, A>.tailRecM(a, f)
+    }
+}
+
+public class FutureKApplicativeError<Err> : FutureKApplicative<Err>, ApplicativeError where Err : Error {
```
raulraja: By convention after 0.8.0 Arrow will always have the type class as first type in the extends clause

truizlop: It is not possible in this case as `FutureKApplicative<Err>` is a (super)class, not a protocol. Therefore, in the inheritance clause it must precede any protocol (typeclass in this case).

raulraja:
🎉

PR #58, opened 2018-11-08
Improve testing for typeclasses laws

truizlop:
Previous version used constant functions in property-based tests for typeclasses laws. This PR replaces those constant functions (where possible) by `ArrowOf<A, B>`, provided by SwiftCheck, which generates functions with signature `(A) -> B`, making the tests of properties more robust.

### DIFF
--- a/Tests/BowTests/Typeclasses/MonadLaws.swift
+++ b/Tests/BowTests/Typeclasses/MonadLaws.swift
@@ -17,10 +17,10 @@ class MonadLaws<F> {
     }
     
     private static func leftIdentity<Mon, EqF>(_ monad : Mon, _ eq : EqF) where Mon : Monad, Mon.F == F, EqF : Eq, EqF.A == Kind<F, Int> {
-        property("Monad left identity") <- forAll { (a : Int, b : Int) in
-            let f = { (_ : Int) in monad.pure(a) }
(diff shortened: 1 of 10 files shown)


Tests/BowTests/Typeclasses/MonadLaws.swift
```diff
@@ -17,10 +17,10 @@ class MonadLaws<F> {
     }
     
     private static func leftIdentity<Mon, EqF>(_ monad : Mon, _ eq : EqF) where Mon : Monad, Mon.F == F, EqF : Eq, EqF.A == Kind<F, Int> {
-        property("Monad left identity") <- forAll { (a : Int, b : Int) in
-            let f = { (_ : Int) in monad.pure(a) }
-            return eq.eqv(monad.flatMap(monad.pure(b), f),
-                          f(b))
+        property("Monad left identity") <- forAll { (a : Int, f : ArrowOf<Int, Int>) in
+            let g = f.getArrow >>> monad.pure
```
JorgeCastilloPrz: What's that operator?

truizlop: `andThen`, reverse of `<<<` or `compose`. Some other libraries already use it.

truizlop:
I usually leave a space in both sides of the `:` in the declarations. It could be a good idea to use a linter as you mention.

PR #60, opened 2018-11-16
Split Bow in multiple targets

truizlop:
This PR introduces the following targets:

- Bow
- BowFree
- BowGeneric
- BowOptics
- BowRecursionSchemes
- BowResult
- BowEffects
- BowBrightFutures
- BowRx

The goal is to have independent modules that can be consumed separatedly. The core library is more lightweight and has no external dependencies. Corresponding `.podspec` files have been created for each target, with their corresponding external dependencies. Regarding Carthage support, it involves downloading the whole project, but then consumers may decide to import only the `.framework` files they need.


.travis.yml
```diff
@@ -3,20 +3,29 @@ osx_image: xcode10
...
-    - env: JOB="XCODE" DEST="OS=12.0,name=Apple TV" SCHEME="Bow-tvOS" SDK="appletvsimulator" ACTION="build"
-    - env: JOB="XCODE" DEST="OS=5.0,name=Apple Watch Series 4 - 44mm" SCHEME="Bow-watchOS" SDK="watchsimulator" ACTION="build"
+    - env: SCHEME="Bow"
+    - env: SCHEME="Bow-Free"
+    - env: SCHEME="Bow-Generic"
+    - env: SCHEME="Bow-Optics"
+    - env: SCHEME="Bow-RecursionSchemes"
+    - env: SCHEME="Bow-Result"
+    - env: SCHEME="Bow-Effects"
+    - env: SCHEME="Bow-Rx"
```
JorgeCastilloPrz: Integrations are named like `arrow-effects-rx`. Is there a reason to avoid the effects prefix here?

truizlop: These are the build scheme names, not the actual module names.

JorgeCastilloPrz: Ok, I thought you had named them like BowRx and so on, as it's stated in the description.

PR #65, opened 2018-11-21
Update README

truizlop:
This PR updates the README to point the badges to the right URLs (they were outdated after the migration to the new organization) and adds information to consume the library from CocoaPods and Carthage.

It also moves some methods from State to a constrained extension of StateT in order to make them more accessible without castings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ pod "BowBrightFutures", "~> 0.2.0"
 
 ### Carthage
 
-Carthage with download the whole Bow project, but it will compile individual frameworks for each module that you can use separately. Add this line to your Cartfile:
+Carthage will download the whole Bow project, but it will compile individual frameworks for each module that you can use separately. Add this line to your Cartfile:
 
 ```
 github "bow-swift/Bow" ~> 0.2.0
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Bow
 
-[![Build Status](https://travis-ci.org/arrow-kt/bow.svg?branch=master)](https://travis-ci.org/arrow-kt/bow)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,7 +20,31 @@ Bow is split in multiple modules that can be consumed independently. These modul
...
+pod "BowGeneric", "~> 0.2.0"
+pod "BowResult", "~> 0.2.0"
+pod "BowEffects", "~> 0.2.0"
+pod "BowRx", "~> 0.2.0"
+pod "BowBrightFutures", "~> 0.2.0"
+```
+
+### Carthage
+
+Carthage with download the whole Bow project, but it will compile individual frameworks for each module that you can use separately. Add this line to your Cartfile:
```
juancazalla: Typo, `Carthage **will** download`

truizlop:
```suggestion
Carthage will download the whole Bow project, but it will compile individual frameworks for each module that you can use separately. Add this line to your Cartfile:
```

README.md
```diff
@@ -20,7 +20,31 @@ Bow is split in multiple modules that can be consumed independently. These modul
 - `BowBrightFutures`: module to provide an integration with BrightFutures.
 - `BowRx`: module to provide an integration with RxSwift.
 
-Bow is available using Cocoapods, Carthage and Swift Package Manager.
+Bow is available using CocoaPods and Carthage.
+
+### CocoaPods
+
+You can consume each Bow module as a separate pod. You can add these lines to your Podfile at your convenience:
```
juancazalla: I would use something like:

```suggestion
You can consume each Bow module as a separate pod. To do it, add these lines to your Podfile at your convenience:
```

Issue #77, opened 2018-12-27
"ap" is backwards?


stephencelis:
Swift can define `ap` in a much more readable (left-to-right, non-nested) format that matches Haskell. By introducing a generic at function scope we can constrain `F`'s `A` to be a function between generics:

``` diff
-public func ap<B>(_ ff: F<(A) -> B>) -> F<B>
+public func ap<A1, B>(_ fa: F<A1>) -> F<B> where A == (A1) -> B
```

And in practice:

``` swift
Option.pure(add).ap(one).ap(two)

// vs. the right-to-left, awkwardly nested:
two.ap(one.ap(Option.pure(add)))
```

FWIW many functional libraries in Swift introduce the `<*>` operator, as another option. That could be an alternate consideration if the inconsistency with Arrow is an issue, though I assume Bow should be optimized for Swift syntax in general.

truizlop:
The definition of `ap` is done to keep consistency with the typeclass definition for `Applicative`, and the behavior of `ap` is not the same as in `<*>`. If we check the signature for both:

```swift
ap : (F<A>, F<(A) -> B>) -> F<B>
<*> : (F<A>, F<B>) -> F<(A) -> B>
```

`<*>` definition, and the usage you are proposing, is encoded in Bow as `product` in the `Applicative` typeclass, and the way you could be using it is:

```swift
Option<Int>.applicative().product(one, two).fix().map(add)
```

The syntax could be better but it is a matter of getting to the point where the language supports higher kinded types natively.

stephencelis:
It may be a language barrier thing, but you have me a bit confused. You don't have `<*>` defined at all in Bow, so there's no signature to compare your `ap` with. You've also mentioned deprecating operators, so I assume `<*>` isn't being added?

In Haskell, `ap` is on the `Monad` typeclass and is defined with the exact same infix way as `<*>` (though with different type class constraints):

``` h
ap :: m (a -> b) -> m a -> m b
<*> :: f (a -> b) -> f a -> f b
```

So is it fair to assume that you want Bow's type classes to be consistent with _Arrow's_? That can definitely be a project prerogative, though porting over a language limitation is unfortunate.

truizlop:
[...]

PR #80, opened 2019-01-03
Refactor ap method in Applicative

truizlop:
As suggested in #77, for a nicer usage of `ap`, the wrapped function should be on the left hand side of `ap`, allowing for a better syntax call. This PR provides this change in every data type supporting `ap`, in the `Applicative` typeclass (swapping arguments for a better curried version) and in all instances of such typeclass.

### DIFF
--- a/Sources/Bow/Data/Const.swift
+++ b/Sources/Bow/Data/Const.swift
@@ -105,7 +105,7 @@ public class ConstApplicative<R, Mono> : ConstFunctor<R>, Applicative where Mono
         return ConstMonoid(self.monoid).empty
     }
     
-    public func ap<A, B>(_ fa: ConstOf<R, A>, _ ff: ConstOf<R, (A) -> B>) -> ConstOf<R, B> {
+    public func ap<A, B>(_ ff: ConstOf<R, (A) -> B>, _ fa: ConstOf<R, A>) -> ConstOf<R, B> {
         return Const.fix(ff).ap(Const.fix(fa), monoid)
     }
 }
--- a/Sources/Bow/Data/Const.swift
+++ b/Sources/Bow/Data/Const.swift
@@ -35,8 +35,8 @@ public class Const<A, T> : ConstOf<A, T> {
         return Const<A, T>(semigroup.combine(self.value, other.value))
     }
     
-    public func ap<AA, U, SemiG>(_ ff : Const<A, AA>, _ semigroup : SemiG) -> Const<A, U> where SemiG : Semigroup, SemiG.A == A, T == (AA) -> U {
-        return self.retag().combine(ff.retag(), semigroup)
+    public func ap<AA, U, SemiG>(_ fa : Const<A, AA>, _ semigroup : SemiG) -> Const<A, U> where SemiG : Semigroup, SemiG.A == A, T == (AA) -> U {
+        return self.retag().combine(fa.retag(), semigroup)
     }
 }
 
--- a/Sources/Bow/Data/Const.swift
+++ b/Sources/Bow/Data/Const.swift
@@ -35,8 +35,8 @@ public class Const<A, T> : ConstOf<A, T> {
         return Const<A, T>(semigroup.combine(self.value, other.value))
     }
     
-    public func ap<U, SemiG>(_ ff : Const<A, (T) -> U>, _ semigroup : SemiG) -> Const<A, U> where SemiG : Semigroup, SemiG.A == A {
-        return ff.retag().combine(self.retag(), semigroup)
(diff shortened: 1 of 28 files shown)


Sources/Bow/Data/Const.swift
```diff
@@ -35,8 +35,8 @@ public class Const<A, T> : ConstOf<A, T> {
         return Const<A, T>(semigroup.combine(self.value, other.value))
     }
     
-    public func ap<U, SemiG>(_ ff : Const<A, (T) -> U>, _ semigroup : SemiG) -> Const<A, U> where SemiG : Semigroup, SemiG.A == A {
-        return ff.retag().combine(self.retag(), semigroup)
+    public func ap<AA, U, SemiG>(_ fa : Const<A, AA>, _ semigroup : SemiG) -> Const<A, U> where SemiG : Semigroup, SemiG.A == A, T == (AA) -> U {
+        return self.retag().combine(fa.retag(), semigroup)
```
miguelangel-dev: In some cases, you add `self` (in the same context - optional) and in the majority, you do not do it. 

Do you think we can unify it, erase them?

Sources/Bow/Data/Either.swift
```diff
@@ -78,8 +78,8 @@ public class Either<A, B> : EitherOf<A, B> {
                     { b in Either<C, D>.right(fb(b)) })
     }
     
-    public func ap<C>(_ ff : Either<A, (B) -> C>) -> Either<A, C> {
-        return ff.flatMap{ f in self.map(f) }
+    public func ap<BB, C>(_ fb : Either<A, BB>) -> Either<A, C> where B == (BB) -> C {
+        return self.flatMap(fb.map)
```
miguelangel-dev:
```suggestion
        return flatMap(fb.map)
```

Sources/Bow/Transformers/EitherT.swift
```diff
@@ -55,8 +55,8 @@ public class EitherT<F, A, B> : EitherTOf<F, A, B> {
         return EitherT<F, A, C>(functor.map(fc, Either<A, C>.right))
     }
     
-    public func ap<C, Mon>(_ ff : EitherT<F, A, (B) -> C>, _ monad : Mon) -> EitherT<F, A, C> where Mon : Monad, Mon.F == F {
-        return ff.flatMap({ f in self.map(f, monad) }, monad)
+    public func ap<BB, C, Mon>(_ fa : EitherT<F, A, BB>, _ monad : Mon) -> EitherT<F, A, C> where Mon : Monad, Mon.F == F, B == (BB) -> C {
+        return self.flatMap({ f in fa.map(f, monad) }, monad)
```
miguelangel-dev:
```suggestion
        return flatMap({ f in fa.map(f, monad) }, monad)
```

Sources/Bow/Transformers/StateT.swift
```diff
@@ -69,8 +69,8 @@ public class StateT<F, S, A> : StateTOf<F, S, A> {
         }.map{ ssz in StateT<F, S, Z>(ssz) }
     }
     
-    public func ap<B, Mon>(_ ff : StateT<F, S, (A) -> B>, _ monad : Mon) -> StateT<F, S, B> where Mon : Monad, Mon.F == F {
-        return ff.map2(self, { f, a in f(a) }, monad)
+    public func ap<AA, B, Mon>(_ fa : StateT<F, S, AA>, _ monad : Mon) -> StateT<F, S, B> where Mon : Monad, Mon.F == F, A == (AA) -> B {
+        return self.map2(fa, { f, a in f(a) }, monad)
```
miguelangel-dev:
```suggestion
        return map2(fa, { f, a in f(a) }, monad)
```

Sources/Bow/Typeclasses/Applicative.swift
```diff
@@ -2,16 +2,16 @@ import Foundation
...
 
 public extension Applicative {
     public func map<A, B>(_ fa: Kind<F, A>, _ f: @escaping (A) -> B) -> Kind<F, B> {
-        return ap(fa, pure(f))
+        return ap(pure(f), fa)
     }
     
     public func product<A, B>(_ fa : Kind<F, A>, _ fb : Kind<F, B>) -> Kind<F, (A, B)> {
-        return self.ap(fb, self.map(fa, { (a : A) in { (b : B) in (a, b) }}))
+        return self.ap(self.map(fa, { (a : A) in { (b : B) in (a, b) }}), fb)
```
miguelangel-dev:
```suggestion
        return ap(self.map(fa, { (a : A) in { (b : B) in (a, b) }}), fb)
```

PR #105, opened 2019-01-16
Read GA id from config

calvellido:
* Simple fix to set the Google Analytics id parameter to be read from config, as it was incorrectly set as a hardcoded value.

### DIFF
--- a/docs/_includes/_google-tag-manager.html
+++ b/docs/_includes/_google-tag-manager.html
@@ -1,5 +1,5 @@
 <!-- Global site tag (gtag.js) - Google Analytics -->
-<script async src="https://www.googletagmanager.com/gtag/js?id=UA-18433785-19"></script>
+<script async src="https://www.googletagmanager.com/gtag/js?id={{ site.google_analytics }}"></script>
 <script>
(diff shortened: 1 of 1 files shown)


docs/_includes/_google-tag-manager.html
```diff
@@ -1,5 +1,5 @@
 <!-- Global site tag (gtag.js) - Google Analytics -->
-<script async src="https://www.googletagmanager.com/gtag/js?id=UA-18433785-19"></script>
+<script async src="https://www.googletagmanager.com/gtag/js?id={{ site.google_analytics }}"></script>
```
truizlop: Where is this `site.google_analytics` defined?

calvellido: Here: https://github.com/bow-swift/bow/blob/99f7f7506011be06da535b1980cebec8ccd31503/docs/_config.yml#L21

calvellido: In Jekyll the way to read values declared in the config file is that way. We could also store it in a different `yml` file though.

PR #136, opened 2019-02-14
New encoding for HKTs and Typeclasses

### LINKED ISSUE #20: Explore how implement typeclasses using extensions
The purpose is to implement typeclasses with less boilerplate.

truizlop:
## Related issues

Closes #20.

## Goal

Current encoding of typeclasses and HKTs make a bit cumbersome to use Bow. Instances of typeclasses need to be manually composed and passed. Also, declaring functions that use a typeclass typically involve writing a long constraint list in a where clause. This is difficult to use, read and maintain.

The goal of this PR is to change the encoding of HKTs and Typeclasses so that the writing of functions using them becomes easier and instances of typeclasses are resolved by the compiler. In the end, we'd like to be able to write code like:

```swift
func divide<F: ApplicativeError>(_ a: Int, _ b: Int) -> Kind<F, Int> where F.E == DivideError {
    guard b != 0 else { return F.raiseError(.zeroDenominator) }
    return F.pure(a/b)
}

print(Either.fix(divide(6, 3).map { a in 2 * a }.map { a in "Str:\(a)" })) // Prints Right(Str:4)
print(Either.fix(divide(6, 0))) // Prints Left(zeroDenominator)
print(Validated.fix(divide(6, 3))) // Prints Valid(2)
print(Validated.fix(divide(6, 0))) // Prints Invalid(zeroDenominator)
```

Here, the `divide` function has a constraint on `F` conforming to `ApplicativeError`. Calling this function fixing the return type takes the correct instance of the typeclass and applies it.

## Implementation details

- Witnesses of simulated HKTs are now final classes. They used to be `typealias`. With this, we can make the witness conform to the typeclass, something that was not possible with an alias before. For instance, we went from:

```swift
public class ForEither {}
public typealias EitherPartial<L> = Kind<ForEither, L>
public typealias EitherOf<L, R> = Kind2<ForEither, L, R>
public class Either<L, R>: EitherOf<L, R> { ... }
```

To:

```swift
public final class ForEither {}
public final class EitherPartial<L>: Kind<ForEither, L>
public typealias EitherOf<L, R> = Kind<EitherPartial<L>, R>
public class Either<L, R>: EitherOf<L, R> { ... }
```

- Typeclasses do not have `F` as an associated type, but refer to `Self`. Also, methods in typeclasses are static. For instance, we went from:

```swift
public protocol Functor {
    associatedtype F
    func map<A, B>(_ fa: Kind<F, A>, _ f: @escaping (A) -> B) -> Kind<F, B>
}
```

To:

```swift
public protocol Functor {
    static func map<A, B>(_ fa: Kind<Self, A>, _ f: @escaping (A) -> B) -> Kind<Self, B>
}
```

- Instances of a typeclass for a given type are extensions on the type witness. For example, the instance of `Functor` for `Either<L, R>` is an extension on its witness `EitherPartial<L>`:

```swift
extension EitherPartial: Functor {
    static func map<A, B>(_ fa: Kind<EitherPartial<L>, A>, _ f: @escaping (A) -> B) -> Kind<EitherPartial<L>, B> {
        // Implementation
     }
}
```

- Syntax is enabled at the `Kind` level via extensions with constraints. This way, methods are available for every object of the type. For example:

```swift
extension Kind where F: Functor {
    func map<B>(_ f: @escaping (A) -> B) -> Kind<F, A> {
        return F.map(self, f)
    }
}
```

Therefore, as `Either` has an instance of Functor, we can call `EitherPartial<String>.map(either, f)`, or `either.map(f)`, making the library syntax much more approachable. Multiple calls to typeclass functions can also be chained: `either.map { x in 2 * x }.map { y in y * y }`. 

- A typeclass `EquatableK` to check equality at the kind level has been introduced. From this, we can automatically derive `Equatable`, to be able to check equality of two kinds using `==`, for those types where it makes sense.

## Testing details

- Laws testing the typeclasses' properties had to be refactored to the new encoding, which dramatically reduced the number of arguments they need to receive to check properties. In most of the cases, it is reduced to a generator function to get values from a given type. This could be automated with generators from SwiftCheck, but it is left as future work.


juancazalla:
AWESOME!

raulraja:
🙌 🙌 🙌 outstanding work, this will highly simplify usage.

PR #199, opened 2019-03-19
Searchbar improvements

### LINKED ISSUE #138: Search bar malfunctioning
## Description

There are some of issues that I have experienced using the search bar:

1) Type `Either`. It does not show on the suggestion list. Hitting enter does not navigate to its docs.
2) Searching from a page different from the main one sometimes leads to 404 errors. An example flow to reproduce it:
- Go to the main page for API docs.
- Type `Lens` in the search bar. Three results `[Lens, Lens, LensPartial]` show.
- Click on the first result and navigate to that page.
- While you are on the `Lens` page, type `Lens` on the search bar again.
- Select the second `Lens` result.
- You should see a 404.

I think this is due to a bad formatting of the URL. Looking at the address bar, it says `https://bow-swift.io/api-docs/Extensions/Typealiases.html#/s:9BowOptics4Lensa`; the `Extensions` in the address bar does not need to be present, but it is because the previous page we visited had it in the path. removing it shows the correct page.

3) If you type an operator like `>>>`, it does not show up in the search results.

calvellido:
## Related issues

#138

## Goal

Fix the problems described in #138.

## Implementation details

These changes fix problems related with some words not being detected, like `Either`, by removing the stopwords filter completely. Problem on navigating and searching again was due to `Extensions` not being detected as a subdirectory for the documentation, it has been now added. Problem regarding some symbols not being detected as search queries was due to a trimmer in the search index, which it's been removed completely now.

It also adds some visual improvements in the way the results are shown, like a small transition effect in the results box appearing, and extending the width to the whole sidebar.

This also updates the Lunr search engine to its version 1.0.0 from 0.7.2.

This fixes #138


calvellido:
Travis was really slow yesterday indeed, I received the build passing notification this dawn :sunrise:  :scream:

truizlop:
Yes, I don't know why... I launched several PRs and each one of them took several hours to complete. Anyway, let's merge this one!

PR #218, opened 2019-03-22
Rename file to match iOS naming convention

Tylos:
## Goal

Follow iOS naming conventions and unify `NonEmptyArray` name

## Implementation details

Rename file from `NonEmptyList+Optics.swift` to `NonEmptyArray+Optics.swift`


truizlop:
Good catch!

Issue #239, opened 2019-03-28
Port OptionT missing functions


Tylos:
## Description

In order to align Bow and [Arrow](https://github.com/arrow-kt/arrow) API, there are some missing methods that require to be ported

* Check [OptionT](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-extras-data/src/main/kotlin/arrow/data/OptionT.kt) in Arrow repository
* Compare with [OptionT](https://github.com/bow-swift/bow/blob/master/Sources/Bow/Transformers/OptionT.swift) in Bow Repository
* Port the missing functions
* Add the appropriate documentation
* Include tests for the added functions

At the moment of creating this issue the existing differences are:
- [x] toLeft
- [x] toRight

PR #245, opened 2019-03-29
Add missing functions to OptionT

juancazalla:
## Related issues

Closes: #239 

## Goal

Add missing methods compared to Arrow version.

### DIFF
--- a/Sources/Bow/Transformers/OptionT.swift
+++ b/Sources/Bow/Transformers/OptionT.swift
@@ -100,8 +100,8 @@ extension OptionT where F: Functor {
     ///
     /// - Parameter defaultRigth: Function returning a default value to use as right if the `OptionT` is none.
     /// - Returns: Returns: An `EitherT` containing the value as left or as right with the default value if the `OptionT` contains a none.
-    public func toLeft<R>(_ defaultRigth: @escaping () -> R) -> EitherT<F, A, R> {
-        return EitherT(cata({ .right(defaultRigth()) }, Either.left))
+    public func toLeft<R>(_ defaultRight: @escaping () -> R) -> EitherT<F, A, R> {
+        return EitherT(cata({ .right(defaultRight()) }, Either.left))
     }
 
     /// Convert this `OptionT` to an `EitherT`.
--- a/Sources/Bow/Transformers/OptionT.swift
+++ b/Sources/Bow/Transformers/OptionT.swift
@@ -103,6 +103,14 @@ extension OptionT where F: Functor {
     public func toLeft<R>(_ defaultRigth: @escaping () -> R) -> EitherT<F, A, R> {
         return EitherT(cata({ .right(defaultRigth()) }, Either.left))
     }
+
+    /// Convert this `OptionT` to an `EitherT`.
+    ///
+    /// - Parameter defaultLeft: Function returning a default value to use as left if the `OptionT` is none.
+    /// - Returns: Returns: An `EitherT` containing the value as right or as left with the default value if the `OptionT` contains a none.
+    public func toRight<L>(_ defaultLeft: @escaping () -> L) -> EitherT<F, L, A> {
+        return EitherT(cata({ .left(defaultLeft()) }, Either.right))
+    }
 }
 
 // MARK: Functions for `OptionT` when the effect has an instance of `Applicative`
--- a/Sources/Bow/Transformers/OptionT.swift
+++ b/Sources/Bow/Transformers/OptionT.swift
@@ -95,6 +95,14 @@ extension OptionT where F: Functor {
     public func subflatMap<B>(_ f: @escaping (A) -> Option<B>) -> OptionT<F, B> {
         return transform { option in Option.fix(option.flatMap(f)) }
     }
+
+    /// Convert this `OptionT` to an `EitherT`.
+    ///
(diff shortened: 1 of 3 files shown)


Sources/Bow/Transformers/OptionT.swift
```diff
@@ -95,6 +95,22 @@ extension OptionT where F: Functor {
     public func subflatMap<B>(_ f: @escaping (A) -> Option<B>) -> OptionT<F, B> {
         return transform { option in Option.fix(option.flatMap(f)) }
     }
+
+    /// Convert this `OptionT` to an `EitherT`.
+    ///
+    /// - Parameter defaultRigth: Function returning a default value to use as right if the `OptionT` is none.
+    /// - Returns: Returns: An `EitherT` containing the value as left or as right with the default value if the `OptionT` contains a none.
+    public func toLeft<R>(_ defaultRigth: @escaping () -> R) -> EitherT<F, A, R> {
+        return EitherT(cata({ .right(defaultRigth()) }, Either.left))
```
truizlop: There is a spelling mistake in `Right`.
```suggestion
        return EitherT(cata({ .right(defaultRight()) }, Either.left))
```

Sources/Bow/Transformers/OptionT.swift
```diff
@@ -95,6 +95,22 @@ extension OptionT where F: Functor {
     public func subflatMap<B>(_ f: @escaping (A) -> Option<B>) -> OptionT<F, B> {
         return transform { option in Option.fix(option.flatMap(f)) }
     }
+
+    /// Convert this `OptionT` to an `EitherT`.
+    ///
+    /// - Parameter defaultRigth: Function returning a default value to use as right if the `OptionT` is none.
+    /// - Returns: Returns: An `EitherT` containing the value as left or as right with the default value if the `OptionT` contains a none.
+    public func toLeft<R>(_ defaultRigth: @escaping () -> R) -> EitherT<F, A, R> {
```
truizlop:
```suggestion
    public func toLeft<R>(_ defaultRight: @escaping () -> R) -> EitherT<F, A, R> {
```

Sources/Bow/Transformers/OptionT.swift
truizlop: You can also replace `{ .right(defaultRight()) }` by `constant(.right(defaultRight()))`, but no big deal, as you prefer.

Sources/Bow/Transformers/OptionT.swift
truizlop: Also, it would be nice to add `@autoclosure` to be able to pass a single value.

Sources/Bow/Transformers/OptionT.swift
```diff
@@ -95,6 +95,22 @@ extension OptionT where F: Functor {
...
+        return EitherT(cata({ .right(defaultRigth()) }, Either.left))
+    }
+
+    /// Convert this `OptionT` to an `EitherT`.
+    ///
+    /// - Parameter defaultLeft: Function returning a default value to use as left if the `OptionT` is none.
+    /// - Returns: Returns: An `EitherT` containing the value as right or as left with the default value if the `OptionT` contains a none.
+    public func toRight<L>(_ defaultLeft: @escaping () -> L) -> EitherT<F, L, A> {
+        return EitherT(cata({ .left(defaultLeft()) }, Either.right))
```
truizlop: Same comments about `constant` and `@autoclosure` as above.

Tests/BowTests/Transformers/OptionTTest.swift
```diff
@@ -40,4 +40,32 @@ class OptionTTest: XCTestCase {
     func testFunctorFilterLaws() {
         FunctorFilterLaws<OptionTPartial<ForId>>.check(generator: self.generator)
     }
+
+    func testToLeftWithSome() {
+        property("toLeft for .some should build a correct EitherT") <- forAll { (a: Int, b: String) in
+            let optionT = OptionT<ForId, Int>.fromOption(.some(a))
+            return optionT.toLeft({ b }) == EitherT<ForId, Int, String>.left(a)
```
truizlop: With `@autoclosure` you don't need the braces in `{ b }`.

juancazalla: Do we prefer the @autoclosure syntax? I mean, in case we want to use a function as a parameter we are going to need `()`.

Sources/Bow/Transformers/OptionT.swift
juancazalla: Using `@autoclosure` seems a good idea if we are fine with the syntax change for the caller, because we can still use `function()` as a parameter and it will not be executed, however (correct me if I'm wrong) using `constant` in the implementation makes me feel that we are losing the advantage of using a function, because we are going to execute the `.right(defaultRight())` and `defaultRight` could be a time-consuming function or could contain a side effect.

Let me know what you think. In fact, I would change `constant` to be implemented by using @autoclosure, in this way we can get the lambda without executing the `constant`.

juancazalla: Take into account that we are not using `@autoclosure` in the library right now, we have a lot of cases that are similar to this one and we are not using it. I'm not against the decision of using it, I just want to verify we want it.

PR #246, opened 2019-03-29
Add missing functions to Try

### LINKED ISSUE #225: Port Try missing functions
## Description

In order to align Bow and [Arrow](https://github.com/arrow-kt/arrow) API there are some missing methods that require to be ported

* Check [Try](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Try.kt) in Arrow repository
* Compare with [Try](https://github.com/bow-swift/bow/blob/master/Sources/Bow/Data/Try.swift) in Bow Repository
* Port the missing functions
* Add the appropriate documentation
* Include tests for the added functions

At the moment of creating this issue the existing differences are:
- [x] mapFilter
- [x] isFailure
- [x] isSuccess
- [x] exists
- [x] toOption
- [x] toEither
- [x] getOrDefault
- [x] orNull
- [x] orElse
- [x] rescue
- [x] flatten

### LINKED ISSUE #233: Create Try missing instances
## Description

In order to align Bow and [Arrow](https://github.com/arrow-kt/arrow) API, there are some missing instances that require to be ported

* Check [Try](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/try.kt) instances in Arrow repository
* Check [Try](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/try.kt) MTL instances in Arrow repository
* Compare with [Try](https://github.com/bow-swift/bow/blob/master/Sources/Bow/Data/Try.swift) instances in Bow Repository
* Create the missing instances for the existing typeclasses
* Add the appropriate documentation
* Include test laws for the created instances

At the moment of creating this issue the missing instances are:
- [x] TrySemigroup
- [x] TryMonoid
- [ ] TryHashable
- [ ] TryFunctorFilter

truizlop:
## Related issues

- Closes #225 
- Closes #233 

## Goal

Add missing functionality to Try respect to the Arrow version.

## Implementation details

Some missing functions and instances have been added. Some notes:

- Instance to `Hashable` has not been added. All instances of `Hashable` should be added separately in a single PR.
- `rescue` is deprecated and replaced by `handleErrorWith` from `ApplicativeError`, which is already present.
- `flatten` is already present from `Monad`.
- `exists` is already present from `Foldable`.

## Testing details

Laws for Semigroup and Monoid are tested.

### DIFF
--- a/Sources/Bow/Data/Try.swift
+++ b/Sources/Bow/Data/Try.swift
@@ -123,7 +123,7 @@ public postfix func ^<A>(_ fa: TryOf<A>) -> Try<A> {
     return Try.fix(fa)
 }
 
-class Success<A>: Try<A> {
+private class Success<A>: Try<A> {
     fileprivate let value: A
     
     init(_ value: A) {
@@ -131,7 +131,7 @@ class Success<A>: Try<A> {
     }
 }
 
-class Failure<A>: Try<A> {
+private class Failure<A>: Try<A> {
     fileprivate let error: Error
     
     init(_ error: Error) {
--- a/Sources/Bow/Data/Try.swift
+++ b/Sources/Bow/Data/Try.swift
@@ -7,7 +7,7 @@ import Foundation
 /// - unsupportedOperation: The invoked operation is unsupported for the value receiving it.
 public enum TryError: Error {
     case illegalState
-    case predicateError
+    case predicateDoesNotMatch
     case unsupportedOperation(String)
 }
 
--- a/Sources/Bow/Data/Try.swift
+++ b/Sources/Bow/Data/Try.swift
@@ -18,13 +18,15 @@ public final class ForTry {}
 public typealias TryOf<A> = Kind<ForTry, A>
 
 /// Describes the result of an operation that may have thrown errors or succeeded. The type parameter corresponds to the result type of the operation.
-public class Try<A>: TryOf<A> {
+public final class Try<A>: TryOf<A> {
+    private let value: _Try<A>
+
     /// Creates a successful Try value.
     ///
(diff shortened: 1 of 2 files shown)


Sources/Bow/Data/Try.swift
```diff
@@ -3,9 +3,11 @@ import Foundation
 /// Models an error in an invocation to some methods of `Try`.
 ///
 /// - illegalState: The value is in an illegal state when an operation is invoked.
+/// - predicateDoesNotMatch: The value does not match the provided predicate.
```
miguelangel-dev:
```suggestion
/// - predicateError: The value does not match the provided predicate.
```

truizlop: I think the current name is more specific to the error and the suggestion is too generic.

miguelangel-dev: I agree, but then I think we should change the `case` option, wdut?

Sources/Bow/Data/Try.swift
```diff
@@ -69,19 +77,27 @@ public class Try<A>: TryOf<A> {
...
-            case let failure as Failure<A>: return fe(failure.error)
-            case let success as Success<A>:
-                do {
-                    return try fa(success.value)
-                } catch let error {
-                    return fe(error)
-                }
-            default:
-                fatalError("Try must only have Success or Failure cases")
+        switch value {
```
miguelangel-dev: I like this change :) good improvement

PR #316, opened 2019-04-15
Docs for FP concepts

truizlop:
## Goal

Provide some documentation for FP core concepts that are addressed in the library.

## Implementation details

Pages for the following content have been written:
- Functions vs Procedures: includes notions on what is FP, what is a function and benefits of programming with them.
- Higher Kinded Types: includes definitions and motivation, how they are implemented in Bow, and how you can make your own HKT.
- Type classes: includes definitions, how they are used, and the core ones provided in Bow.
- Data types: includes definitions, how they are used, and the core ones provided in Bow.
- Glossary: acts as a dictionary of recurring terms that are used across the docs.

### DIFF
--- a/contents/Documentation/FP concepts.playground/Pages/Data types.xcplaygroundpage/Contents.swift
+++ b/contents/Documentation/FP concepts.playground/Pages/Data types.xcplaygroundpage/Contents.swift
@@ -54,5 +54,5 @@
  | OptionT<F, A> | Represents an `Option<A>` nested in an arbitrary effect `F`.|
  | StateT<F, S, A> | Represents a function that receives an effect of type `S` and produces a new state of the same type and an output value of type `A`, all wrapped in an effect of type `F`.|
  | State<S, A> | Represents a StateT where the effect is Id.|
- | WriterT<F, W, A> | Represents a value of type `A` under the effect `F` that produdes a side stream of data of type `W`.|
+ | WriterT<F, W, A> | Represents a value of type `A` under the effect `F` that produces a side stream of data of type `W`.|
  */
--- a/contents/Documentation/FP concepts.playground/Pages/Data types.xcplaygroundpage/Contents.swift
+++ b/contents/Documentation/FP concepts.playground/Pages/Data types.xcplaygroundpage/Contents.swift
@@ -1,0 +1,58 @@
+// nef:begin:header
+/*
+ layout: docs
+ title: Data types
+ */
+// nef:end
+/*:
+ # Data types
+
+ {:beginner}
+ beginner
+
+ A data type is an abstraction that encapsulates a reusable coding pattern. Data types are generalized by one or more type parameters. Data types in Bow have Higher Kinded Type support by extending `Kind`. Their functionality works on their own structure, never on the values defined by its generic type parameters.
+
+ ## Data types in Bow
+
+ The following tables summarize some of the data types included in Bow, together with a short description of their purpose.
+
+ ### Function-like types
+
+ | Data type | Purpose |
+ | --------- | ------- |
+ | Function0<A> | Represents a constant function; one that does not receive any parameters and produces a constant value of type `A`.|
+ | Function1<I, O> | Represents a function that receives values of type `I` and produces values of type `O`.|
+ | Kleisli<F, D, A> | Represents a function that receives values of type `D` and produces values of type `A` wrapped in an effect `F` (namely, `Kind<F, A>`). Dual of Cokleisli.|
+ | ReaderT<F, D, A> | Equivalent to `Kleisli<F, D, A>`.|
+ | Reader<D, A> | Represents a ReaderT where the effect is Id.|
+ | Cokleisli<F, D, A> | Represents a function that receives values of type `A` wrapped in an effect `F` (namely, `Kind<F, A>`), and produces values of type `D`. Dual of Kleisli.|
+ | CoreaderT<F, D, A> | Equivalent to `Cokleisli<F, D, A>`.|
+ | Coreader<F, D, A> | Represents a CoreaderT where the effect is Id.|
+
+ ### Core types
+
+ | Data type | Purpose |
+ | --------- | ------- |
+ | ArrayK<A> | Represents an array of values of type `A`. It is like `Array<A>` or `[A]`, but with HKT support.|
+ | Const<A, T> | Represents a constant value of type `A` and a phantom type `T`.|
+ | DictionaryK<K, A> | Represents a dictionary where keys have type `K` and values have type `A`. It is like `Dictionary<K, A>` or `[K: A]`, but with HKT support.|
+ | Either<A, B> | Represents the sum type of `A` and `B`; i.e., it holds a value of either one of those types.|
+ | Eval<A> | Represents a potentially lazy value of type `A`. |
+ | Id<A> | Represents a value of type `A` with no further context.|
+ | Ior<A, B> | Represents a value of either `A` or `B`, or both values at the same time.|
+ | NonEmptyArray<A> | Represents an array of one or more elements of type `A`.|
+ | Option<A> | Represents a value of type `A` that may or may not be present. It is like `Optional<A>` or `A?`, but with HKT support.|
+ | SetK<A> | Represents an unordered collection of unique values of type `A`. It is like `Set<A>` but with HKT support. |
+ | Try<A> | Represents a computation that may have provided a value of type `A` or thrown an error.|
+ | Validated<A, B> | Represents a value that may be invalid, with an error value of type `A`, or a valid value of type `B`.
+
+ ### Transformers
+
+ | Data type | Purpose |
+ | --------- | ------- |
+ | EitherT<F, A, B> | Represents an `Either<A, B>` nested in an arbitrary effect `F`.|
+ | OptionT<F, A> | Represents an `Option<A>` nested in an arbitrary effect `F`.|
+ | StateT<F, S, A> | Represents a function that receives an effect of type `S` and produces a new state of the same type and an output value of type `A`, all wrapped in an effect of type `F`.|
+ | State<S, A> | Represents a StateT where the effect is Id.|
+ | WriterT<F, W, A> | Represents a value of type `A` under the effect `F` that produdes a side stream of data of type `W`.|
+ */
--- a/contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
+++ b/contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
@@ -168,7 +168,7 @@ let identity2 = id <<< f1
 
  ## Working with impure functions
 
- Real world software rarely has functions as nicely written as the `add` function above. We usually have to deal with partial, non-deterministic and/or impure functions, since software has to deal with errors and perform side effects in order to be useful. Besides presenting issues breaking referential transparency, function composition cannot be done in this circumstances. How can we deal with such situations?
+ Real world software rarely has functions as nicely written as the `add` function above. We usually have to deal with partial, non-deterministic and/or impure functions, since software has to deal with errors and perform side effects in order to be useful. Besides presenting issues breaking referential transparency, function composition cannot be done in these circumstances. How can we deal with such situations?
 
  Bow provides numerous data types that can be used to model different effects in our functions. Using those data types, such as `Option`, `Either` or `IO` can help us transform partial, non-deterministic or impure functions into total, deterministic and pure ones. Once we have them, we can use combinators that will help us compose those functions nicely.
  */
--- a/contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
+++ b/contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
@@ -153,7 +153,7 @@ let composed4 = f1 >>> f2
  Composition of functions is associative; i.e. the following functions are equivalent:
  */
 func f3(_ x: String) -> String {
-    return x.rev
+    return String(x.reversed())
 }
 
 let associativity1 = (f3 <<< f2) <<< f1
--- a/contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
+++ b/contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
@@ -4,11 +4,171 @@
  title: Functions vs. Procedures
  */
 // nef:end
+// nef:begin:hidden
+import Bow
+// nef:end
 /*:
  # Functions vs. Procedures
 
  {:.beginner}
  beginner
  
- Content under development.
+ According to Wikipedia, **Functional Programming** is a programming paradigm - a style of building the structure and elements of computer programs - that treats computation as the evaluation of mathematical functions and avoids changing state and mutable data. In short, we can say that Functional Programming is programming with **functions**.
+
+ Functions, as they are understood in FP, must have the following properties:
+
+ - **Total**: functions must provide an output for every input.
+ - **Deterministic**: for a given input, the function always returns the same output.
+ - **Pure**: the evaluation of the function does not cause any other effects besides computing the output.
+
+ We can notice that this does not totally match with the definition of functions (or methods) as entities declared in Swift using the `func` keyword, as the language does not enforce any of the previous properties when we are creating our functions. When we create a function that does not match any of the following properties stated above, we can refer to it as a **procedure**. Sometimes you can also read them as **pure functions** (matching the 3 properties above) versus **impure functions** (procedures).
+
+ Let us look at the following code:
+ */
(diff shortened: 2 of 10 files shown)


contents/Documentation/FP concepts.playground/Pages/Data types.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,58 @@
...
+
+ ### Transformers
+
+ | Data type | Purpose |
+ | --------- | ------- |
+ | EitherT<F, A, B> | Represents an `Either<A, B>` nested in an arbitrary effect `F`.|
+ | OptionT<F, A> | Represents an `Option<A>` nested in an arbitrary effect `F`.|
+ | StateT<F, S, A> | Represents a function that receives an effect of type `S` and produces a new state of the same type and an output value of type `A`, all wrapped in an effect of type `F`.|
+ | State<S, A> | Represents a StateT where the effect is Id.|
+ | WriterT<F, W, A> | Represents a value of type `A` under the effect `F` that produdes a side stream of data of type `W`.|
```
juancazalla:
```suggestion
 | WriterT<F, W, A> | Represents a value of type `A` under the effect `F` that produces a side stream of data of type `W`.|
```

contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,171 @@
...
+let impureSquare = impureAdd(x: 2, y: 5) * impureAdd(x: 2, y: 5)
+/*:
+ If we apply the same technique as above to eliminate the common subexpressions:
+ */
+let impureSum = impureAdd(x: 2, y: 5)
+let impureSquare2 = impureSum * impureSum
+/*:
+ The result of the two operations is the same, but the observed behavior of the overall program is not the same: the first implementation prints twice to the console, as there are two invocations to `impureAdd`, whereas in the second case there is only one print when the sum is saved in `impureSum`.
+
+ Referential transparency lets us have a better reasoning of our programs as we do not have to keep in mind the state of the computation and consider any side effects happening after the execution of a certain piece of code. Moreover, code that is referentially transparent is intrinsically testable (we just need to provide inputs and assert over the expected outputs) and can be optiimized and parallelized easily (there is no shared mutable state that causes concurrency issues).
```
juancazalla:
```suggestion
 Referential transparency lets us have a better reasoning of our programs as we do not have to keep in mind the state of the computation and consider any side effects happening after the execution of a certain piece of code. Moreover, code that is referentially transparent is intrinsically testable (we just need to provide inputs and assert over the expected outputs) and can be optimized and parallelized easily (there is no shared mutable state that causes concurrency issues).
```

contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,171 @@
...
+
+ ## Memoization
+
+ Another consequence of having a function that is referentially transparent is that it can be **memoized**. Memoization is a technique used to cache already computed values, specially if they have a high cost to be computed. When a function is memoized, you can think of it as a lookup table where inputs and their corresponding outputs are saved. Subsequent invocations just retrieve the value from the table instead of actually computing it.
```
juancazalla:
```suggestion
 Another consequence of having a function that is referentially transparent is that it can be **memoized**. Memoization is a technique used to cache already computed values, especially if they have a high cost to be computed. When a function is memoized, you can think of it as a lookup table where inputs and their corresponding outputs are saved. Subsequent invocations just retrieve the value from the table instead of actually computing it.
```

contents/Documentation/FP concepts.playground/Pages/Functions vs Procedures.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,171 @@
...
+ Given three functions, it does not matter if we compose the first two, and then the third one, or the latter two and then the first one. There is also a function that, when composed with any other function, does nothing: the `id` (identity) function, which is included in Bow.
+ */
+let identity1 = f1 <<< id
+let identity2 = id <<< f1
+/*:
+ The functions above have the exact same behavior as `f1`.
+
+ ## Working with impure functions
+
+ Real world software rarely has functions as nicely written as the `add` function above. We usually have to deal with partial, non-deterministic and/or impure functions, since software has to deal with errors and perform side effects in order to be useful. Besides presenting issues breaking referential transparency, function composition cannot be done in this circumstances. How can we deal with such situations?
```
juancazalla:
```suggestion
 Real world software rarely has functions as nicely written as the `add` function above. We usually have to deal with partial, non-deterministic and/or impure functions, since software has to deal with errors and perform side effects in order to be useful. Besides presenting issues breaking referential transparency, function composition cannot be done in these circumstances. How can we deal with such situations?
```

contents/Documentation/FP concepts.playground/Pages/Glossary.xcplaygroundpage/Contents.swift
```diff
@@ -10,5 +10,85 @@
...
+
+ ### Function
+
+ A function is a piece of code that has the following properties:
+
+ - **Total**: it must provide an output for every input.
+ - **Deterministic**: for a given input, the function always returns the same output.
+ - **Pure**: the evaluation of the function does not cause any other effects besides computing the output.
+
+ Notice that Swift does not enforce any of these properties in their definition. Swift functions that have these properties are generally referred as **pure functions**, whereas the ones that do not have some of them are called **impure functions**.
```
juancazalla:
```suggestion
 Notice that Swift does not enforce any of these properties in its definition. Swift functions that have these properties are generally referred to as **pure functions**, whereas the ones that do not have some of them are called **impure functions**.
```

truizlop: `Their` refers to functions here.

contents/Documentation/FP concepts.playground/Pages/Glossary.xcplaygroundpage/Contents.swift
```diff
@@ -10,5 +10,85 @@
...
+
+ ### Memoization
+
+ **Memoization** is an optimization technique that transforms expensive computational function calls into searches in a lookup table by caching previous calls.
+
+ # P
+
+ ### Parametric polymorphism
+
+ **Parametric polymorphism** is a type of polymorphism that allows to specify generic types to create functions and/or other types. These generic types are called *type parameters*.
```
juancazalla:
```suggestion
 **Parametric polymorphism** is a type of polymorphism that allows specifying generic types to create functions and/or other types. These generic types are called *type parameters*.
```

contents/Documentation/FP concepts.playground/Pages/Glossary.xcplaygroundpage/Contents.swift
```diff
@@ -10,5 +10,85 @@
...
+
+ ### Referential Transparency
+
+ **Referential transparency** is a property of a function that allows it to be replaced by the result of its execution without altering the overall behavior of the program.
+
+ # T
+
+ ### Type
+
+ A **type** is a set of values that helps us conceptualize data and restrict the possible values that functions can accept as input or provide as output. Types can be finite (like `Bool`, with two values: `true` and `false`) or infinite (like `String`).
```
juancazalla:
```suggestion
 A **type** is a set of values that help us conceptualize data and restrict the possible values that functions can accept as input or provide as output. Types can be finite (like `Bool`, with two values: `true` and `false`) or infinite (like `String`).
```

truizlop: set -> helps

contents/Documentation/FP concepts.playground/Pages/Higher Kinded Types.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,193 @@
...
+ ```swift
+    func divide<F: ErrorSuccessRepresentable>(x: Int, y: Int) -> F<DivideError, Int> {
+        guard y != 0 else { return .failure(.divisionByZero)
+        return .success(x / y)
+    }
+ ```
+
+ That is, given a type `F` that is able to create success and error values (via the hypothetical `ErrorSuccessRepresentable` protocol), we could write a function that checks if the division can be performed or not, and create the result values accordingly. This would allow us to write generic programs where we can generalize the container types, not only the contained ones.
+
+ Unfortunately, Swift does not support writting the code above since it does not have HKTs support. Can we find a workaround to have this feature?
```
juancazalla:
```suggestion
 Unfortunately, Swift does not support writing the code above since it does not have HKTs support. Can we find a workaround to have this feature?
```

contents/Documentation/FP concepts.playground/Pages/Higher Kinded Types.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,193 @@
...
+
+ ## Types and Kinds
+
+ Before we answer this question, let us review some basic concepts. The notion of **type** is well-known among software engineers and Swift developers. Being a class, struct or enum, a type is a set of values. This set may be finite, as in the case of `Bool` where there are only two values (`true` and `false`), or infinite, as in the case of `String`. Types group values that are similar. If we raise the level of abstraction, we could group similar types in sets of types; those *sets of types* are **kinds**.
```
juancazalla:
```suggestion
 Before we answer this question, let us review some basic concepts. The notion of **type** is well-known among software engineers and Swift developers. Being a class, struct or enum, a type is a set of values. This set may be finite, as in the case of `Bool` where there are only two values (`true` and `false`), or infinite, as in the case of `String`. Types, group values that are similar. If we raise the level of abstraction, we could group similar types in sets of types; those *sets of types* are **kinds**.
```

Not sure about this one

contents/Documentation/FP concepts.playground/Pages/Type classes.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,144 @@
...
+func allEqual<A: Equatable>(_ array: [A]) -> Bool {
+    guard let first = array.first else { return false }
+    return array.reduce(true) { partial, next in partial && next == first }
+}
+/*:
+ The function above is parametrically polymorphic; it operates of values of type `A`. Valid types to fill this type parameter need to conform to `Equatable` and enable the posibility to use the `==` operator.
+
+ ## Type classes of higher kind
+
+ `Equatable`, as many other protocols provided by Foundation or any other Swift framework, operates at the type level. However, there are many other interesting abstractions as type classes that need to operate at higher kind level. This is not possible by default in Swift, but Bow simulates HKTs and enable type classes to work at this level.
```
juancazalla:
```suggestion
 `Equatable`, like many other protocols provided by Foundation or any other Swift framework, operates at the type level. However, there are many other interesting abstractions as type classes that need to operate at higher kind level. This is not possible by default in Swift, but Bow simulates HKTs and enables type classes to work at this level.
```

juancazalla: Two changes in the previous suggestion

contents/Documentation/FP concepts.playground/Pages/Type classes.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,144 @@
...
+ - Transitivity: if `a == b` and `b == c`, then `a == c`.
+
+ The `Equatable` type class can be used to constrain implementations of parametrically polymorphic functions; e.g:
+ */
+func allEqual<A: Equatable>(_ array: [A]) -> Bool {
+    guard let first = array.first else { return false }
+    return array.reduce(true) { partial, next in partial && next == first }
+}
+/*:
+ The function above is parametrically polymorphic; it operates of values of type `A`. Valid types to fill this type parameter need to conform to `Equatable` and enable the posibility to use the `==` operator.
```
juancazalla:
```suggestion
 The function above is parametrically polymorphic; it operates of values of type `A`. Valid types to fill this type parameter need to conform to `Equatable` and enable the possibility to use the `==` operator.
```

contents/Documentation/FP concepts.playground/Pages/Higher Kinded Types.xcplaygroundpage/Contents.swift
truizlop: group acts as a verb and the subject is types, they cannot be separated by a comma.

contents/Documentation/FP concepts.playground/Pages/Type classes.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,144 @@
...
+ - **Composition**: if we apply two transformations to a value, we must obtain the same result as if we apply a single transformation with the composition of the two transforming functions; i.e., `F.transform(F.transform(fa, f), g) == F.transform(fa, g <<< f)`.
+
+ Swift does not provide a way of encoding these laws and enforcing them in every implementation. To do this, Bow type classes encode their laws as property-based tests using [SwiftCheck](https://github.com/typelift/SwiftCheck). Then, each implementation can be tested against these laws to guarantee that the implementation satisfies them. Type class laws help us have a better reasoning about our code and some times help us rewrite pieces for optimization (like in the case of `Transformer` where we can compose functions and apply a single transformation instead of two).
+
+ Notice that sometimes you can provide an implementation for a given type class that does not satisfies the laws. Those are called **lawless instances**, in contrast to the ones that satisfy them, which are called **lawful instances**.
```
juancazalla:
```suggestion
 Notice that sometimes you can provide an implementation for a given type class that does not satisfy the laws. Those are called **lawless instances**, in contrast to the ones that satisfy them, which are called **lawful instances**.
```

contents/Documentation/FP concepts.playground/Pages/Type classes.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,144 @@
...
+}
+// nef:end
+extension ForOption: Transformer {
+    static func transform<A, B>(_ fa: Kind<ForOption, A>, _ f: (A) -> B) -> Kind<ForOption, B> {
+        return fa^.fold(Option<B>.none,                // It is empty, no transformation
+                        { a in Option<B>.some(f(a)) }) // Transformed with f and wrapped in an Option<B>
+    }
+}
+/*:
+ Notice that, since the type class works at the kind level, it is not an extension of `Option`, but and extension of `ForOption`. Revisiting the definition of `Transformer`, the `Self` requirement is used in `Kind<Self, A>`. If we disect the definition of `Option<A>`, it extends `Kind<ForOption, A>`. By doing some sort of pattern matching between `Kind<Self, A>` and `Kind<ForOption, A>`, we can see that it must be `ForOption` the candidate to do the extension.
```
juancazalla:
```suggestion
 Notice that, since the type class works at the kind level, it is not an extension of `Option`, but an extension of `ForOption`. Revisiting the definition of `Transformer`, the `Self` requirement is used in `Kind<Self, A>`. If we dissect the definition of `Option<A>`, it extends `Kind<ForOption, A>`. By doing some sort of pattern matching between `Kind<Self, A>` and `Kind<ForOption, A>`, we can see that it must be `ForOption` the candidate to do the extension.
```

contents/Documentation/FP concepts.playground/Pages/Glossary.xcplaygroundpage/Contents.swift
juancazalla: 👍

contents/Documentation/FP concepts.playground/Pages/Glossary.xcplaygroundpage/Contents.swift
juancazalla: Ops, you cheating! haha, yep, you are right

contents/Documentation/FP concepts.playground/Pages/Higher Kinded Types.xcplaygroundpage/Contents.swift
```diff
@@ -4,11 +4,193 @@
...
+/*:
+ The operator becomes much more convenient when chaining methods. For instance, the `swap` method is defined in `Either` and not available in `Kind`; therefore, we need to cast:
+ */
+let swapped = either.map { value in value.description }^.swap() // swapped is of type Either<String, DivideError>
+/*:
+ ## HKT support for existing types
+
+ Previous sections have covered how to create a new type with HKT support. However, there are a number of types that are already created and probably outside our control. How do we add HKT support for them?
+
+ Unfortunately, there is no way to do this out of the box using the mechanisms that Swift provides for extension. The solution that we have adopted is to wrap the existing type into another type with HKT support, and delegate methods to the internal wrapped value.
```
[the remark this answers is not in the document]
juancazalla: I think that both are valid here, whatever you prefer.

PR #321, opened 2019-04-17
Fix wrong nef tags

truizlop:
## Goal

Remove open nef tag without matching closing tag.

### DIFF
--- a/contents/Documentation/FP concepts.playground/Pages/Higher Kinded Types.xcplaygroundpage/Contents.swift
+++ b/contents/Documentation/FP concepts.playground/Pages/Higher Kinded Types.xcplaygroundpage/Contents.swift
@@ -12,13 +12,13 @@ import Bow
 
  {:.beginner}
  beginner
- 
+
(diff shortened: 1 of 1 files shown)


contents/Documentation/FP concepts.playground/Pages/Higher Kinded Types.xcplaygroundpage/Contents.swift
```diff
@@ -57,10 +57,10 @@ func divideValidated(x: Int, y: Int) -> Validated<DivideError, Int> {
  We can see some similarities between these two functions. They are checking that the second argument is not 0 to perform the division and wrap it in a right/valid case. Otherwise, the division cannot be performed and they wrap an error in a left/invalid case. It would be helpful to unify these two functions in a similar way as in the `allEqual` example presented above. Hypothetically, we would like to write something like:
 
  ```swift
-    func divide<F: ErrorSuccessRepresentable>(x: Int, y: Int) -> F<DivideError, Int> {
-        guard y != 0 else { return .failure(.divisionByZero)
-        return .success(x / y)
-    }
+ func divide<F: ErrorSuccessRepresentable>(x: Int, y: Int) -> F<DivideError, Int> {
```
miguelangel-dev: just curiosity - why do we delete the indentation?

truizlop: That's strange, I didn't do that... since it is in the comments, Xcode must have re-indented when I opened the file. I'll fix that later since this is already merged.

PR #333, opened 2019-05-21
add @escaping keyword to runAsync<A> methods of Async and IO

Dragna:
## Goal

let a user call the callback defined in the runAsync PR, previous to this PR, the compiler would complain about the callback being non-escaping (Closure use of non-escaping parameter 'callback' may allow it to escape)

## Implementation details

add @escaping to the runAsync method in IO.swift and Async.swift

## Testing details

if you want me to add tests, where do you want me to put those ? :)


Dragna:
You are right,  I didn't think about the other implementations of Async 🙂.

It's done.

truizlop:
Thank you very much @Dragna!

PR #336, opened 2019-06-06
Property-based generators for types in the core module

truizlop:
## Goal

Create generators for Property-based Testing for the types in the core module so that their values are used in the tests for laws. For this purpose, SwiftCheck is used.

## Implementation details

- Some types (mostly the ones that do not have nested kinds) have been made to conform `Arbitrary`.
- A new type class `ArbitraryK` has been introduced to be able to generate values at the kind level and abstract over this functionality.
- A final wrapper type `KindOf` is introduced. `Arbitrary` is derived for this type if the contained `F` conforms to `ArbitraryK`.
- All types in the core module have conformance to `ArbitraryK`.

## Testing details

Future work will make use of these generators to test the type class laws for the instances of the corresponding types.


Sources/Bow/Data/Either.swift
```diff
@@ -204,20 +206,9 @@ public postfix func ^<A, B>(_ fa: EitherOf<A, B>) -> Either<A, B> {
...
-    }
-}
-
-private class Right<A, B> : Either<A, B> {
-    let b : B
-    
-    init(_ b : B) {
-        self.b = b
-    }
+private enum _Either<A, B> {
```
miguelangel-dev: Good improvement!

Sources/Bow/Data/Ior.swift
```diff
@@ -178,30 +183,10 @@ public postfix func ^<A, B>(_ fa: IorOf<A, B>) -> Ior<A, B> {
...
-
-class IorBoth<A, B>: Ior<A, B> {
-    fileprivate let a: A
-    fileprivate let b: B
-    
-    init(_ a: A, _ b: B) {
-        self.a = a
-        self.b = b
-    }
+private enum _Ior<A, B> {
```
miguelangel-dev: definitively it is a nice change

Tests/BowGenerators/Data/Sum+Gen.swift
```diff
@@ -0,0 +1,14 @@
+import Bow
+import SwiftCheck
+
+// MARK: Instance of `ArbitraryK` for `Sum`
+
+extension SumPartial: ArbitraryK where F: ArbitraryK, G: ArbitraryK {
```
miguelangel-dev: review this proposal: https://github.com/bow-swift/bow/pull/337

PR #338, opened 2019-06-10
Generators for Effects, Rx, BrightFutures and Free

truizlop:
## Goal

Add generators for other types outside the core module.

## Implementation details

Based on the work done in #336, we extend it to create PBT generators for Effects, Rx, BrightFutures and Free.


miguelangel-dev:
LGTM! :)

Issue #315, opened 2019-04-12
Docs versioning


truizlop:
## Description

Current website for Bow only shows the state of the docs for the current state of the code base. As we move forward, we need to enable versioning of the docs for users to be able to retrieve the correct docs for the version they are using.

PR #339, opened 2019-06-11
Docs versioning

calvellido:
## Related issues

#315 

## Goal

The goal is to have different Bow docs versions built and available to consult on our website.

## Implementation details

The different versions are read from the Git tags present in the project. This is done through a Ruby script, which run in a Travis environment, and the script supports different options that can be set, e.g. the tags for which we want docs to be generated from.

---

This only sets the CI to build the different version microsites under a structure suitable to be used moving forward. The version selector is being habdled through another PR (#334)


scripts/gen-docs.rb
```diff
@@ -1,21 +1,115 @@
...
+# Generate the Jekyll site through Nef based on the contents source.
+#
+# @param version [String] The version for which the nef docs site will be generated.
+# @return [nil] nil.
+def generate_nef_site(version)
+  system "echo Generating nef site for #{version}"
+  # system "mkdir -p docs/#{version}/_data"
+  system "nef jekyll --project contents/Documentation --output docs --main-page contents/Home.md"
+  # system "nef jekyll --project contents/Documentation --output docs/#{version} --main-page contents/Home.md"
+  system "JEKYLL_ENV=production BUNDLE_GEMFILE=./docs/Gemfile bundle exec jekyll build -s ./docs -d ./docs/#{version} -b bow/#{version}"
```
miguelangel-dev: uh! smart

calvellido:
I'm merging this. For now, the default version (the one appearing at [bow-swift.io/docs](https://bow-swift.io/docs)) is set to show `master` branch content, as the `0.4.0` tag wasn't ready to be served this way yet.

Once we have a new version released, we can set the script to build the latest version by default, just setting:

```diff
- default_version = "master"
+ default_version = ""
```

PR #340, opened 2019-06-12
Update laws to use generators for Property-based Testing

truizlop:
## Goal

Leveraging the work done in #336 and #338, laws for the core module type classes are rewritten to use the generators for the given types. As a consequence of having a more general input, some bugs in certain implementations have arisen and been fixed.


miguelangel-dev:
I love how the code has been clean up with the new generators `ArbitratiK` — also, good catch using the same generator instance.

Just a simple question, what is the dependency who forces us to import 
```
 Bow
 BowGenerators
 BowRxGenerators
```
instead to just using the `testable` import?

PR #341, opened 2019-06-12
Automatic derivation of optics

truizlop:
## Goal

The usage of the current Optics module is hindered by the need to write all optics for our data types. This PR adds mechanisms to automatically obtain optics for product and sum types. In particular, the ones that we are able to generate at the moment are Lenses and Optionals, and with a little bit of boilerplate, Prisms.

## Implementation details

The implementation of this feature leverages the introduction of key paths in Swift, which allows us to abstract over fields and let us set/get their values without referencing them explicitly. Protocols `AutoLens`, `AutoOptional` and `AutoPrism` have been introduced to enable this feature in new values.

### Lenses

As an example, consider the following data structures:

```swift
struct Article: AutoLens, AutoOptional {
    var title: String
    var date: Date
    var author: Person
    var keyword: Option<String>
}

struct Person: AutoLens {
    var name: String
    var age: Int
}
```

Notice that fields need to be marked as `var` since the `AutoLens` or `AutoOptional` protocols need to be able to create copies of them and set the new values. One can then access lenses for each field:

```swift
let titleLens: Lens<Article, String> = Article.lens(for: \.title)
let ageLens: Lens<Person, Int> = Person.lens(for: \.age)
```

We can also access nested fields by composing lenses:

```swift
let authorNameLens = Article.lens(for: \.author) + Person.lens(for: \.name)
```

Or even better:

```swift
let authorNameLens = Article.lens(for: \.author.name)
```

### Optionals

Similarly, we can also get `Optional` optics for fields that are defined as `A?` or `Option<A>`:

```swift
let keywordOptional = Article.optional(for: \.keyword)
```

### Prisms

Finally, we can do a similar thing with `Prism` although it involves a bit more boilerplate as sum types do not have syntax similar to key paths or generalized pattern matching. Consider the following example:

```swift
enum GameState: AutoPrism {
    case notStarted
    case inProgress(Board)
    case draw(Board)
    case won(Board, winner: Stone)
}
```

We can get a prism focusing on a case that do not have any associated value like:

```swift
let notStartedPrism: Prism<GameState, ()> = GameState.prism(for: GameState.notStarted)
```

If the case has associated values, then a matching function needs to be provided:

```swift
let inProgressPrism: Prism<GameState, Board> = GameState.prism(for: GameState.inProgress) { state in
    switch state {
    case .inProgress(let board): return board
    default: return nil
    }
}

let wonPrism: Prism<GameState, (Board, Stone)> = GameState.prism(for: GameState.won) { state in
    switch state {
    case .won(let board, let stone): return (board, stone)
    default: return nil
    }
}
```

The matching closure must only match the case of interest to the prism and return a tuple of its associated values, or nil if the match does not succeeds.

### DIFF
--- a/Bow.xcodeproj/project.pbxproj
+++ b/Bow.xcodeproj/project.pbxproj
@@ -488,6 +488,13 @@
 			remoteGlobalIDString = 1120978222A7EBC7007F3D9C;
 			remoteInfo = "Bow-Generators";
 		};
+		1126CA8022B0DDD100F840CB /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = OBJ_1 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 117DC0F722AE522000EF65F0;
+			remoteInfo = "Bow-BrightFuturesGenerators";
+		};
 		117DC0B422AE492700EF65F0 /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
 			containerPortal = OBJ_1 /* Project object */;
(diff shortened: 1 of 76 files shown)


nomisRev:
Very nice solution for Swift 👏 👏 👏

PR #342, opened 2019-06-12
Lenses for Tuples

truizlop:
## Goal

After the work done in #341, it is very likely to have a `Prism` targeting a tuple. However, there are no utilities at the moment to manipulate components of a tuple with optics. This PR adds lenses for every component of tuples of arities from 2 to 10, thus enhancing the composition of different optics.


miguelangel-dev:
Great!

PR #343, opened 2019-06-13
Permalinks trailing slash additions

calvellido:
## Goal

I'm not sure if this is the way to proceed, or that information is being read from another place through `nef`, but when writing permalinks for Jekyll is better to always write them ending at `/` slash. That way the files/path generated in the end could be navigated through `bow-swift.io/docs` and `bow-swift.io/docs/`. Without it, a path not ending in a slash will give you a 404 😵

Do you think there could be more elsewhere?  Thanks!

### DIFF
--- a/contents/Documentation/FP concepts.playground/docs/docs/README.md
+++ b/contents/Documentation/FP concepts.playground/docs/docs/README.md
@@ -1,7 +1,7 @@
 ---
 layout: docs
 title: Home
-permalink: /docs
+permalink: /docs/
 ---
 
 # Bow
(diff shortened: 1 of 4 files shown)


truizlop:
I think this comes from `nef`, so perhaps the fix needs to be done there. Can you confirm that @miguelangel-dev?

miguelangel-dev:
Yep! We need to modify nef. is it an standard in Jekyll?

miguelangel-dev:
Moved this issue to `nef` repo - https://github.com/bow-swift/nef/issues/61

miguelangel-dev:
I have checked, and `nef` is working on this way. It generates the permalink finished `/`

miguelangel-dev:
I will prepare a different PR fixing static content - just in `main page`

PR #374, opened 2019-06-25
Add Resource and new UnitDurations

### LINKED ISSUE #264: Duration
## Description

Add utilities to represent `Duration` as per [its definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Duration.kt). Explore if Swift already provides a standard way to represent this.

## Modules

- `BowEffects`.

### LINKED ISSUE #270: Resource
## Description

Create `Resource` as per its [definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/Resource.kt).

## Modules

- `BowEffects`.

truizlop:
## Related issues

- Closes #264 
- Closes #270 

## Goal

- Implements new time units as an extension of `UnitDuration`.
- Implements `Resource` class to handle acquiring and releasing resources properly.

### DIFF
--- a/Sources/BowEffects/Data/Resource.swift
+++ b/Sources/BowEffects/Data/Resource.swift
@@ -9,11 +9,11 @@ public class Resource<F, E, A>: ResourceOf<F, E, A> where F: Bracket, F.E == E {
         return value as! Resource<F, E, A>
     }
     
-    public static func from(_ acquire: @escaping () -> Kind<F, A>, _ release: @escaping (A, ExitCase<E>) -> Kind<F, ()>) -> Resource<F, E, A> {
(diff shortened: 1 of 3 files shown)


Sources/BowEffects/Data/Resource.swift
```diff
@@ -0,0 +1,103 @@
+import Bow
+
+public final class ForResource {}
```
miguelangel-dev: I think we should go adding in the same PR the API docs. It is easier done at that moment :)

truizlop: That would be ideal, yes, but this is prone to changes before merging it to master, so I prefer to make sure everything is stable before adding the API reference. But definitely, it must be documented before merging it with master.

miguelangel-dev:
Sure! I understand :)

Issue #285, opened 2019-04-01
IO


truizlop:
## Description

Update `IO` to match the [definition in Arrow], but maintaining the explicit error type parameter. Some of its missing methods are:

- [x] defer
- [ ] runAsyncCancellable
- [ ] unsafeRunAsyncCancellable
- [x] unsafeRunSync
- [ ] unsafeRunTimed
- [ ] uncancelable
- [x] bracket
- [x] bracketCase
- [x] guarantee
- [x] guaranteeCase

## Modules

- `BowEffects`.

PR #377, opened 2019-06-26
Add Promise

### LINKED ISSUE #268: Promise
## Description

Create `Promise` as per its [definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/Promise.kt).

- [x] Implement `Promise`
- [x] Implement `CancelablePromise`
- [x] Implement `UnsafePromise`

## Modules

- `BowEffects`.

truizlop:
## Related issues

- Closes #268 

## Goal

Add `Promise` concurrency primitive. There are two concrete implementations: `CancelablePromise` and `UncancelablePromise`. There is an unsafe implementation `UnsafePromise`. A protocol `PromiseError` to describe errors that can be thrown from a `Promise` is also included.

### DIFF
--- a/Sources/BowEffects/Data/Atomic.swift
+++ b/Sources/BowEffects/Data/Atomic.swift
@@ -11,17 +11,14 @@ final class Atomic<A> {
         get {
             return queue.sync { self._value }
         }
+        set {
+            self._value = newValue
+        }
     }
 
     func mutate(_ transform: (inout A) -> ()) {
         queue.sync {
             transform(&self._value)
-        }
-    }
-    
-    func set(_ newValue: A) {
-        queue.sync {
-            self._value = newValue
         }
     }
 
--- a/Sources/BowEffects/Data/Internal/CancelablePromise.swift
+++ b/Sources/BowEffects/Data/Internal/CancelablePromise.swift
@@ -1,0 +1,5 @@
+import Bow
+
+internal class CancelablePromise<F: Concurrent, A>: Promise<F, A> {
+    
+}
--- a/Sources/BowEffects/Data/Internal/CancelablePromise.swift
+++ b/Sources/BowEffects/Data/Internal/CancelablePromise.swift
@@ -60,36 +60,27 @@ internal class CancelablePromise<F: Concurrent, A: Equatable>: Promise<F, A> whe
     }
     
     private func unsafeTryComplete(_ a: A) -> Kind<F, Bool> {
-        let current = state.value
-        guard case let .pending(joiners: joiners) = current else {
-            return F.pure(false)
-        }
-        
-        if state.compare(current, andSet: .complete(value: a)) {
-            if !joiners.values.isEmpty {
-                return callAll(joiners.arrayValues, with: .right(a)).map { _ in true }
-            } else {
-                return F.pure(true)
-            }
-        } else {
-            return unsafeTryComplete(a)
-        }
+        return unsafeTry(.complete(value: a), .right(a))
     }
     
     private func unsafeTryError(_ error: F.E) -> Kind<F, Bool> {
+        return unsafeTry(.error(error), .left(error))
+    }
+    
+    private func unsafeTry(_ newState: PromiseState<F.E, A>, _ result: Either<F.E, A>) -> Kind<F, Bool> {
         let current = state.value
         guard case let .pending(joiners: joiners) = current else {
             return F.pure(false)
         }
         
-        if state.compare(current, andSet: .error(error)) {
+        if state.compare(current, andSet: newState) {
             if !joiners.values.isEmpty {
-                return callAll(joiners.arrayValues, with: .left(error)).map { _ in true }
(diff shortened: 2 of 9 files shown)


Sources/BowEffects/Data/Atomic.swift
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    func mutate(_ transform: (inout A) -> ()) {
+        queue.sync {
+            transform(&self._value)
+        }
+    }
+    
+    func set(_ newValue: A) {
```
miguelangel-dev: you can use the same `var value: A` to do this 
```
var value: A {
  get { .... }
  set { queue.sync { ... } }
}
```

Sources/BowEffects/Data/Internal/CancelablePromise.swift
```diff
@@ -0,0 +1,156 @@
...
+    override func tryError<E: Error>(_ e: E) -> Kind<F, Bool> {
+        if let error = e as? F.E {
+            return F.defer { self.unsafeTryError(error) }
+        } else {
+            return F.pure(false)
+        }
+    }
+    
+    private func unsafeTryComplete(_ a: A) -> Kind<F, Bool> {
+        let current = state.value
```
miguelangel-dev: **just a suggestion** I think if you use a `guard case let .pending(joiners: joiners) = state.value else { F.pure(false); return }` it is more readable :)

Sources/BowEffects/Data/Internal/CancelablePromise.swift
```diff
@@ -0,0 +1,156 @@
...
+            } else {
+                return unsafeTryComplete(a)
+            }
+        default:
+            return F.pure(false)
+        }
+    }
+    
+    private func unsafeTryError(_ error: F.E) -> Kind<F, Bool> {
+        let current = state.value
```
miguelangel-dev: same suggestion...

Sources/BowEffects/Data/Internal/Dictionary+Extensions.swift
```diff
@@ -0,0 +1,19 @@
...
+
+func +<K: Hashable, V>(lhs: Dictionary<K, V>, rhs: (K, V)) -> Dictionary<K, V> {
+    var copy = lhs
+    copy[rhs.0] = rhs.1
+    return copy
+}
+
+extension Dictionary {
+    var arrayValues: [Value] {
+        return self.map { x in x.value }
```
miguelangel-dev: Apple gives you something similar... self.values -> `Array(values)`

Sources/BowEffects/Data/Internal/Token.swift
```diff
@@ -0,0 +1,23 @@
...
+extension Token: Hashable {
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(id)
+    }
+}
+
+extension Token: Equatable {}
+
+func ==(lhs: Token, rhs: Token) -> Bool {
+    return lhs === rhs
```
miguelangel-dev: maybe, in this case, it is better to check just the `id` - for example, if we cached it or similar? - I want to say if exists a case where changing the object (persist/retrieve, copy...) and do not change the value (id)

Sources/BowEffects/Data/IO.swift
```diff
@@ -284,17 +284,10 @@ private class AsyncIO<E: Error, A> : IO<E, A> {
...
-        } catch let error as E {
-            result = RaiseError<E, A>(error)
-        } catch {
-            fatalError("IO did not handle error: \(error). Only errors of type \(E.self) are handled.")
+        let callback: Callback<E, A> = { either in
+            result = either.fold(RaiseError.init, Pure.init)
         }
+        f(callback)
         
         while(result == nil) {}
```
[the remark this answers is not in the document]
truizlop: Yes, the ideal would be a `DispatchGroup`. Since `IO` will be refactored to conform to the new type class hierarchy, I'd leave this out for the moment. Issue #285 tracks this.

Sources/BowEffects/Data/Internal/Token.swift
truizlop: It is a reference value and the intention is to check they are actually the same object, not with the same id.

Sources/BowEffects/Data/Internal/CancelablePromise.swift
truizlop: Nice catch!

Sources/BowEffects/Data/Internal/Dictionary+Extensions.swift
truizlop: Replaced.

Sources/BowEffects/Data/IO.swift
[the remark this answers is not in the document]
miguelangel-dev: Great!

PR #390, opened 2019-07-10
Revamp IO

### LINKED ISSUE #285: IO

### LINKED ISSUE #288: IOConnection
## Description

Add `IOConnection` as per its [definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IOConnection.kt).

## Modules

- `BowEffects`.

### LINKED ISSUE #290: IOFrame
## Description

Add `IOFrame` as per its [definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IOFrame.kt).

## Modules

- `BowEffects`.

### LINKED ISSUE #291: IORacePair
## Description

Add `IORacePair` as per its [definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORacePair.kt).

## Modules

- `BowEffects`.

### LINKED ISSUE #292: IORaceTriple
## Description

Add `IORaceTriple` as per its [definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORaceTriple.kt).

## Modules

- `BowEffects`.

### LINKED ISSUE #293: IOStart
## Description

Add `IOStart` as per its [definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IOStart.kt).

## Modules

- `BowEffects`.

### LINKED ISSUE #294: IORunLoop
## Description

Add `IORunLoop` as per its [definition in Arrow](https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt).

## Modules

- `BowEffects`.

truizlop:
## Related issues

- Closes #285
- Closes #288
- Closes #290
- Closes #291 
- Closes #292 
- Closes #293 
- Closes #294 

## Goal

Revamp IO with the new type classes for effects.

## Implementation details

As in the previous version of IO, this data type has two type arguments to be explicit about the errors that it is able to handle. `IO<E, A>` requires that `E` conforms to `Error` to be able to throw errors when they are run unsafely.

There are two type aliases though for common cases. One describes an IO that cannot fail (`E == Never`) and the other describes an IO handling any error (`E == Error`). Those are named `UIO<A> = IO<Never, A>` and `Task<A> = IO<Error, A>`. Their corresponding partials are also added as `ForUIO` and `ForTask`.

`Bracket` has been separated from the hierarchy of the type classes for effects.

`Concurrent` has been modified and it does not support cancellation so far. It only allows parallel execution of different tasks.

`ConcurrentEffect` does not have an instance for `IO` at the moment since cancellation is not implemented yet.

Some concurrency primitives have been deleted in favor of using the ones provided by the Grand Central Dispatch. Internal implementation of IO also uses `DispatchQueue` for running asynchronous tasks and `DispatchGroup` for synchronization of different jobs.


miguelangel-dev:
very nice work!

PR #434, opened 2019-09-23
Effects usage documentation

truizlop:
## Goal

This PR adds documentation pages under a new entry called `Effects` that describes how to work with effects in Bow.

- **Overview**: describes the Bow Effects module and the artifacts that are included.
- **Suspending side effects**: describes how to create an `IO` or `EnvIO` from a side-effectful operation.
- **Manipulating side effects**: describes how to use the functions from the `Functor` hierarchy on an `IO` value.
- **Handling errors**: describes how to raise errors in an `IO` context and how to map and handle them.
- **Concurrency**: describes how to switch execution context and running things in parallel.
- **Running side effects**: describes how to execute an `IO` synchronous and asynchronously.
- **Handling resources**: describes how to acquire-use-release resources using `Bracket`.
- **Foundation extensions**: describes which operations have been added to Foundation to increase compatibility with Bow Effects.


contents/Documentation/Effects.playground/Pages/Suspending side effects.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+/*:
+ The functional versions above have an additional benefit: they are explicit about the type of the errors that may happen during their execution.
+ 
+ Besides `invoke`, `IO` has other methods to create a suspended side effect, like the `invokeEither` above, `invokeResult`, `invokeValidated` and `invokeTry`.
+ 
+ ## Creating an IO from an asynchronous call
+ 
+ Let's look now at the `homePage` function. It provides a generic (non-typed) error or a `Data` value, but it does it through a callback. How can an asynchronous call can be suspended into an `IO` value?
```
juancazalla:
```suggestion
 Let's look now at the `homePage` function. It provides a generic (non-typed) error or a `Data` value, but it does it through a callback. How can an asynchronous call be suspended into an `IO` value?
```

contents/Documentation/Effects.playground/Pages/Running side effects.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,71 @@
...
+enum APIError: Error {}
+// nef:end
+func fetchArticles(from category: Category, page: UInt, limit: UInt) -> IO<APIError, [Article]>
+// nef:begin:hidden
+{ return IO.pure([])^ }
+// nef:end
+/*:
+ ## Synchronous run
+ 
+ We can run the function above synchronously using the `unsafePerformSync` method on `IO`:
```
miguelangel-dev:
```suggestion
 We can run the function above synchronously using the `unsafeRunSync ` method on `IO`:
```

contents/Documentation/Effects.playground/Pages/Suspending side effects.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,189 @@
...
+ 
+ ## Creating an IO from a synchronous call
+ 
+ Dealing with this type of functions is usual in our programming routine, either because we have some non-functional legacy code or because we are using libraries that do not provide a functional API. What can we do to overcome this issue?
+ 
+ Bow Effects provides the IO data type. It is a data type that lets us suspend the execution of side effects, providing us a value that describes the side effect, but it has not been performed yet. That way, we can convert our programs into values that we can combine and compose.
+ 
+ `IO` has a method called `invoke` that can help us suspend a side effect. For instance, the `greet` function above can be rewritten as:
+ */
+func greetIO(name: String) -> IO<Never, Void> {
```
miguelangel-dev: what do you think to change it to `Task<Void` in order to show another IO options

truizlop: I thought about it but it can be a bit confusing for newcomers. I wanted to only use `IO` in the entire documentation and let the aliases as enhancements that more experienced users can use.

PR #443, opened 2019-10-01
Add Semigroup and Monoid instances for Dictionary

truizlop:
## Goal

Add Semigroup and Monoid instances for Dictionary and DictionaryK

### DIFF
--- a/Sources/Bow/Data/Dictionary.swift
+++ b/Sources/Bow/Data/Dictionary.swift
@@ -1,0 +1,21 @@
+import Foundation
+
+// MARK: Instance of `Semigroup` for `Dictionary`
+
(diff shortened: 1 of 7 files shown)


Sources/Bow/Data/Dictionary.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+// MARK: Instance of `Semigroup` for `Dictionary`
+
+extension Dictionary: Semigroup {
+    public func combine(_ other: Dictionary<Key, Value>) -> Dictionary<Key, Value> {
+        var copy = self
+        other.forEach { key, value in
```
juancazalla: Can't we use `merging`?

PR #457, opened 2019-10-06
Use fastlane to create github releases

devesh-shetty:
## Related issues
https://github.com/bow-swift/bow/issues/36

## Goal
This PR automates the following steps in the release process using fastlane:

- Adds fastlane setup.
- Update the `version` in `Bow.podsec`.
- Create a tag with message added in the `CHANGELOG` file.

## Implementation details
I have added a new `CHANGELOG` file. We could include any release notes into that file and our lane will automatically include that in the release note when creating the tag.

## Testing details
- Run `bundle update`
- Run `bundle exec fastlane release version_number:` with a supplied version and see the release being created along with update to `Bow.podsec` file


fastlane/Fastfile
```diff
@@ -0,0 +1,25 @@
...
+
+# Uncomment the line if you want fastlane to automatically update itself
+# update_fastlane
+
+default_platform(:ios)
+
+platform :ios do
+  desc "Prepare for release"
+  lane :release  do |options|
+    version_bump_podspec(path: "Bow.podspec", version_number: options[:version_number])
```
devesh-shetty: @miguelangel-dev I know you mentioned updating podsec file is a manual step but I can do it via this lane using the version number. Do you think that would be a problem?

fastlane/Fastfile
```diff
@@ -0,0 +1,25 @@
...
+    add_git_tag(tag: options[:version_number], message: (File.read("../CHANGELOG") rescue "No changelog provided"))
+    push_git_tags
```
devesh-shetty: `pod lib lint` is failing at the moment on master. I'll look into it for a fix and then integrate `Propagate to CocoaPod the Bow modules using the podspecs`

fastlane/Fastfile
truizlop: That should not be a problem. However, I think we would need to bump the other podspecs in the repo as well. There are many other modules and we keep them in sync with the same version number.

fastlane/Fastfile
truizlop: Which pods are failing? During release of version 0.6.0 I had to remove support for watchOS and tvOS from the Effects module and the ones depending on it, which may be related to this issue. We need to bring back that support, but I can create a separate issue to track it.

devesh-shetty:
> Besides that, do you think we can automate the `pod trunk push` step as well? If so, I can provide the order in which podspecs need to be pushed to satisfy the dependencies.

Yes, please provide me with the order

fastlane/Fastfile
devesh-shetty: When I run ‘pod lib lint’ inside the root directory it fails around swift versioning, I haven’t looked much into it, I’ll do that

truizlop:
The order of publication should be:

- Bow
- BowEffects
- BowOptics
- BowRx
- BowFree
- BowGeneric
- BowRecursionSchemes
- BowGenerators
- BowEffectsGenerators
- BowRxGenerators
- BowFreeGenerators
- BowLaws
- BowEffectsLaws
- BowRxLaws

During the last release we did, I noticed that the trunk repo in Cocoapods was taking a while to refresh, so pushing a podspec that depends on another one was failing during several minutes until the repo somehow processes it and knows a new version is available.

fastlane/Fastfile
devesh-shetty: Ah so the issue was I had a version of cocoapods which did not support `swift_versions`. I have added cocoapods version to Gemfile at the root directory to prevent others from facing a similar issue. `pod lib lint` ran successfully.

fastlane/Fastfile
```diff
@@ -0,0 +1,55 @@
...
+  "BowGeneric.podspec",
+  "BowRecursionSchemes.podspec",
+  "BowGenerators.podspec",
+  "BowEffectsGenerators.podspec",
+  "BowRxGenerators.podspec",
+  "BowFreeGenerators.podspec",
+  "BowLaws.podspec",
+  "BowEffectsLaws.podspec",
+  "BowOpticsLaws.podspec",
+].freeze
```
devesh-shetty: @truizlop I added the files in the order you mentioned

truizlop: Thanks!

PR #458, opened 2019-10-08
Fix bad space char at docs

calvellido:
## Goal

An incorrect space char is causing a title in the _Running Side Effects_ page to not be properly styled: 

https://bow-swift.io/next/docs/effects/running-side-effects/

If the character gets analyzed it's the standard Unicode space (https://unicode.org/cldr/utility/character.jsp?a=0020), but somehow it is not, as it can be checked in this commit diff...

:man_shrugging: :man_shrugging: :man_shrugging: 

https://bow-swift.io/next/docs/effects/running-side-effects/

## Implementation details

This PR set a proper space on that title.

### DIFF
--- a/contents/Documentation/Effects.playground/Pages/Running side effects.xcplaygroundpage/Contents.swift
+++ b/contents/Documentation/Effects.playground/Pages/Running side effects.xcplaygroundpage/Contents.swift
@@ -33,7 +33,7 @@ func fetchArticles(from category: Category, page: UInt, limit: UInt) -> IO<APIEr
 { return IO.pure([])^ }
 // nef:end
 /*:
- ## Synchronous run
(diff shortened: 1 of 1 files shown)


miguelangel-dev:
Good catch! :)

calvellido:
Just curious, was it this @truizlop?

https://twitter.com/tomasruizlopez/status/1181837718000738304

miguelangel-dev:
@calvellido exactly, it was the bug you fixed here; it is related to Apple IDE, as Xcode Playgrounds as Xcode IDE when you tap Option + Space, IDE inserts this wrong blank space instead to regular space.

calvellido:
Good to know, thanks!

PR #460, opened 2019-10-09
Add Semiring typeclass

### LINKED ISSUE #448: Add Semiring type class
## Description

Add `Semiring` type class together with its laws.

Links in Arrow:
- [Semiring](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Semiring.kt)
- [SemiringLaws](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemiringLaws.kt)

## Modules

- Bow (core)

iboss-ptk:
Resolve #448

Also add `Semiring` instances for `Bool` and Numbers

For `Bool`, I also update `Monoid` instance that define `&&` and `true`
as `Monoid` which is more multiplicative. So I change to `||` and
`false` for `Monoid` and add `&&` and `true` to `Semiring`
multiplication

### DIFF
--- a/Sources/Bow/Instances/NumberInstances.swift
+++ b/Sources/Bow/Instances/NumberInstances.swift
@@ -264,7 +264,18 @@ extension Float: Monoid {
     }
 }
 
-/// Instance of `Semigroup` for `Double`. Uses sum as combination of elements.
+/// Instance of `Semiring` for `Float`. Uses multiplication as multiplication of elements and one as one element.
+extension Float: Semiring {
+    public func multiply(_ other: Float) -> Float {
+        return self * other
+    }
+    
+    public static func one() -> Float {
+        return 1
+    }
+}
+
+/// Instance of `Semigroup` for `Double`. Uses addition as combination of elements.
 extension Double: Semigroup {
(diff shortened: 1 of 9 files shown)


Sources/Bow/Instances/NumberInstances.swift
```diff
@@ -14,6 +14,17 @@ extension Int: Monoid {
     }
 }
 
+/// Instance of `Semiring` for `Int`. Uses multiplication as multiplication of elements and one as one element.
```
truizlop:
```suggestion
/// Instance of `Semiring` for `Int`. Uses product (`*`) as multiplication of elements and `1` as one element.
```

truizlop: Can you change the rest of the comments accordingly?

iboss-ptk: @traviscibot updated :) and yes I have run all the tests.

truizlop:
Fantastic job, thanks for your contribution!

PR #463, opened 2019-10-16
Commit package.resolved to the repo

devesh-shetty:
## Goal
This PR commits the `Package.resolved` file generated by `swift build`. It is an auto-generated file. 
According to the [proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0175-package-manager-revised-dependency-resolution.md#proposed-solution),

> The version of every resolved dependency will be recorded in a Package.resolved file in the top-level package, and when this file is present in the top-level package it will be used when performing dependency resolution, rather than the package manager finding the latest eligible version of each package. 

> When this file is checked in, it allows a team to coordinate on what versions of the dependencies they should use. If this file is gitignored, each user will separately choose when to get new versions based on when they run the swift package update command, and new users will start with the latest eligible version of each dependency.

By including this file in the repo, this also ensures everyone is resolving the same version of the dependencies. Additionally, we can determine the same for CI.

## Implementation details
Nothing special over here. It is an auto-generated file. We can update our packages to the latest version by running `swift package update`.

## Testing details
Green build

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,61 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "FileCheck",
+        "repositoryURL": "https://github.com/llvm-swift/FileCheck.git",
+        "state": {
+          "branch": null,
+          "revision": "00945abac615e5203701be0f090ed0e2d2ff15bf",
+          "version": "0.2.0"
+        }
+      },
+      {
+        "package": "Rainbow",
+        "repositoryURL": "https://github.com/onevcat/Rainbow.git",
+        "state": {
+          "branch": null,
+          "revision": "9c52c1952e9b2305d4507cf473392ac2d7c9b155",
+          "version": "3.1.5"
+        }
+      },
+      {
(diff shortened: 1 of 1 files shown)


devesh-shetty:
@truizlop seems like this needs to be checked in. Not sure if it was not checked in before for any particular reason

truizlop:
We have had issues in the past when this file was checked in. Xcode was not able to properly resolve dependencies when the file was present. I guess that is a bug in Xcode... not sure how to proceed with this.

devesh-shetty:
> We have had issues in the past when this file was checked in. Xcode was not able to properly resolve dependencies when the file was present. I guess that is a bug in Xcode... not sure how to proceed with this.

Not sure if this is still an issue on the stable versions of Xcode. We have checked in this file at my org and have had no problems for the past month. We have already included `Gemfile.lock` which is kind of doing the same thing for other dependencies.

truizlop:
Let's try to merge it and we can revert if we observe any issues.

PR #467, opened 2019-10-22
Fix project

truizlop:
## Related issues

After last PRs were merged, some of the files were not properly added to their corresponding targets, causing build failures when the project is build from Xcode. It does not happen on CI since they are still placed in the correct folder and the SPM is able to find them.

## Implementation details

- Adds files to the corresponding targets.
- Deletes a reference to a non-existing file.


devesh-shetty:
Should we also include the xcodebuild step in CI to assure it properly builds?

truizlop:
I'll add the steps for xcodebuild in another PR.

Issue #236, opened 2019-03-28
Port Day missing functions


Tylos:
## Description

In order to align Bow and [Arrow](https://github.com/arrow-kt/arrow) API, there are some missing methods that require to be ported

* Check [Day](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-extras-data/src/main/kotlin/arrow/data/Day.kt) in Arrow repository
* Compare with [Day](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-extras-data/src/main/kotlin/arrow/data/Day.kt) in Bow Repository
* Port the missing functions
* Add the appropriate documentation
* Include tests for the added functions

At the moment of creating this issue the existing differences are:
- [ ] mapLazy
- [ ] ap
- [ ] apLazy
- [ ] coflatMapLazy

devesh-shetty:
Could you assign this to me next? Thanks!
cc: @truizlop @miguelangel-dev

PR #470, opened 2019-11-02
Add Co, Pairing (a few Rank2 types)

bkase:
## Related issues

This work should help with #236 , though I'm not aware of a specific issue for Co and Pairing. I'm creating this PR as suggested here: https://twitter.com/tomasruizlopez/status/1190692278421086209

## Goal

Add Co and Pairing types to Bow as well as introduce the first implementation of Rank2 types that fully works. As implemented right now, Day's functions that should have a higher-rank forall don't work properly.

## Implementation details

I tried _a lot_ of ways to get rank2 typing to work in Swift, and the method here is the one that worked the best.

Here are some of the ways I've attempted that I remember for future reference:

1. Look at Day both in the Arrow and Bow projects and try to do rank2 types in a similar way -- this failed because of an issue with subclassing and generics in Swift and the lack of abstract classes (I believe Arrow's Day is correct, but Kotlin can handle this in a way Swift can't)
2. Since Rank2 foralls are the same as Rank1 existential types, I tried to use protocols with associated types (essentially existential type parameters) to emulate the rank2 foralls. Unfortunately, these are really not ergonomic, and would require redefining the higher-kinded types in Bow using protocols (and it's way less usable than the subclass approach).

So, I went with an approach of referring to the Rank2 variables with `Any`, there is some up-casting needed when constructing these types, but the down-casting is hidden in the implementation.

This was sufficient enough to get everything working.

As an example, I built a simple counter component using these primitives: https://gist.github.com/bkase/f9e338729eb0ef036b4e5b5b81684fc5 in the Phil Freeman Comonadic UI style: https://blog.functorial.com/posts/2016-08-07-Comonads-As-Spaces.html

For further references see 
https://github.com/paf31/purescript-pairing
and https://hackage.haskell.org/package/hdo-0.1/candidate/docs/Pairing.html
and http://hackage.haskell.org/package/kan-extensions-5.2/docs/Control-Monad-Co.html

## Testing details

Pairing's only significant logic is in `zap` and `zap` is used by Co's EquatableK instance, so I didn't feel the need to make a PairingTests.swift file explicitly. Co just runs through the Functor, Applicative, and Monad laws.

### DIFF
--- a/Bow.xcodeproj/project.pbxproj
+++ b/Bow.xcodeproj/project.pbxproj
@@ -7,6 +7,10 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		09CEF84E236E28150070CF43 /* Co.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09CEF84D236E28150070CF43 /* Co.swift */; };
+		09CEF851236E28680070CF43 /* Pairing.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09CEF850236E28680070CF43 /* Pairing.swift */; };
+		09CEF854236E2CFB0070CF43 /* CoTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09CEF853236E2CFB0070CF43 /* CoTest.swift */; };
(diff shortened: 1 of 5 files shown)


bkase:
Oh I also omitted the implementation of `tailRecM` as I didn't need it for my example and it's not super obvious how to write it!

truizlop:
Thanks! I will try to continue from here and implement the tailRecM function.

PR #477, opened 2019-11-27
Extract EnvIO and provide some instances

truizlop:
## Related issues

`EnvIO` was scattered through the `IO` file and starting to grow too big. Also, some instances can be provided for `Kleisli` when `F` has instances for some effect type classes.

## Goal

- Extract `EnvIO` to its own file.
- Provide type class instances for `Kleisli` when `F` has instances for `MonadDefer`, `Async` and `Concurrent`.
- Test laws for the given instances.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
     - bow-swift/nef
     packages:
     - sourcekitten
-    - nef
+    - nef@0.3.2
 
 env:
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -9,7 +9,7 @@ addons:
     - bow-swift/nef
     packages:
     - sourcekitten
-    - nef
+    - nef@0.3.2
```
miguelangel-dev: I think we can wait for "fix Carthage PR" - I have time to review it

PR #479, opened 2019-11-29
Scheduling policies to retry and repeat effectful computations

truizlop:
## Related issues

Current implementation does not have a mechanism to retry or repeat effectful computations based on `IO` or `EnvIO`.

## Goal

Create a composable way of specifying retrial/repeat policies to re-run effectful computations. I used [ZIO](https://github.com/zio/zio/blob/fae35db0dd8c8fb89e52a76bdf2f22d63a81a4e1/core/shared/src/main/scala/zio/Schedule.scala) as an inspiration to achieve this.

## Implementation details

Functions to `retry` and `repeat` have been added to `EnvIO` and `IO` (the latter delegates on the former). These functions receive a `Schedule` that dictates if the effect needs to be reapplied or not.

Primitive `Schedules` have been added, where the most prominent ones are:

- `Schedule.recurs(n)`: recurs `n` times.
- `Schedule.forever()`: recurs forever.
- `Schedule.spaced(time)`: recurs forever, spacing each execution the specified time.
- `Schedule.linear(time)`: recurs forever, incrementing the delay between each attempt in a linear manner, using `time` as a baseline.
- `Schedule.exponential(time)`: recurs forever, incrementing the delay between each attempt in an exponential manner, using `time` as a baseline.
- `Schedule.fibonacci(time)`: recurs forever, incrementing the delay between each attempt following the Fibonacci sequence, using `time` as a baseline.
- `Schedule.duration(time)`: recurs until the specified `time` elapses.

These primitives can be composed in different manners, always resulting in new `Schedule`s that can be subsequently composed. The most prominent composition operators are:

- `s1.and(s2)`: executes as long as both schedules want to keep recursing, using the maximum of both delays.
- `s1.or(s2)`: executes as long as one of the schedules want to keep recursing, using the minimum of both delays.
- `s1.andThen(s2)`: executes `s1` until it states to finish, and then executes `s2`.

## Usage

An example of usage:

Retries the `io` operation 8 times, waiting 2 seconds between each attempt.

```swift
io.retry(Schedule.recurs(8).and(Schedule.spaced(.seconds(2)))
```

### DIFF
--- a/Sources/BowEffects/Data/Internal/DispatchTimeInterval+Extensions.swift
+++ b/Sources/BowEffects/Data/Internal/DispatchTimeInterval+Extensions.swift
@@ -1,0 +1,25 @@
+import Foundation
+
+extension DispatchTimeInterval {
+    func toDouble() -> Double? {
+        switch self {
+        case .seconds(let value): return Double(value)
+        case .milliseconds(let value): return Double(value) * 0.001
+        case .microseconds(let value): return Double(value) * 0.000001
+        case .nanoseconds(let value): return Double(value) * 0.000000001
+        case .never: return nil
+        @unknown default: return nil
+        }
+    }
+}
+
+func >=(lhs: DispatchTimeInterval, rhs: DispatchTimeInterval) -> Bool {
+    let now = DispatchTime.now()
+    return now + lhs >= now + rhs
+}
+
+func -(lhs: DispatchTime, rhs: DispatchTime) -> DispatchTimeInterval {
+    let l = Int(lhs.rawValue)
+    let r = Int(rhs.rawValue)
+    return .nanoseconds(Int(l - r))
+}
--- a/Sources/BowEffects/Data/Internal/DispatchTimeInterval+Extensions.swift
+++ b/Sources/BowEffects/Data/Internal/DispatchTimeInterval+Extensions.swift
@@ -5,8 +5,8 @@ extension DispatchTimeInterval {
         switch self {
         case .seconds(let value): return Double(value)
         case .milliseconds(let value): return Double(value) * 0.001
-        case .microseconds(let value): return Double(value) * 0.000001
-        case .nanoseconds(let value): return Double(value) * 0.000000001
+        case .microseconds(let value): return Double(value) * 0.000_001
+        case .nanoseconds(let value): return Double(value) * 0.000_000_001
         case .never: return nil
         @unknown default: return nil
         }
--- a/Sources/BowEffects/Data/Schedule.swift
+++ b/Sources/BowEffects/Data/Schedule.swift
@@ -70,7 +70,7 @@ public class Schedule<R, A, State, B> {
     ///
     /// - Parameter that: Schedule to be composed with this one.
     public func choose<State2, C>(_ that: Schedule<R, C, State2, B>) -> Schedule<R, Either<A, C>, (State, State2), B> {
-        self.choose(that).map { b in b.fold(id, id) }
+        self.choose(that).map { b in b.merge() }
     }
     
     /// Composes this schedule with another one, providing a new schedule that continues as long as both continue, using the maximum delay of the two schedules.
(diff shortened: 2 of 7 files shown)


Sources/BowEffects/Data/Internal/DispatchTimeInterval+Extensions.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+extension DispatchTimeInterval {
+    func toDouble() -> Double? {
+        switch self {
+        case .seconds(let value): return Double(value)
+        case .milliseconds(let value): return Double(value) * 0.001
+        case .microseconds(let value): return Double(value) * 0.000001
```
miguelangel-dev:
```suggestion
        case .microseconds(let value): return Double(value) * 0.000_001
```

I think it is easier to read

Sources/BowEffects/Data/Internal/DispatchTimeInterval+Extensions.swift
```diff
@@ -0,0 +1,25 @@
...
+        case .nanoseconds(let value): return Double(value) * 0.000000001
```
miguelangel-dev:
```suggestion
        case .nanoseconds(let value): return Double(value) * 0.000_000_001
```

truizlop: Done!

Sources/BowEffects/Data/Schedule.swift
```diff
@@ -0,0 +1,768 @@
...
+                    s2 <- that.update(self.extract(a, s.0), s.1),
+                    yield: (s1.get, s2.get))^
+            }
+        )
+    }
+    
+    /// Pipes the output of the provided schedule as the input of this schedule. Effects described by the provided schedule will always be performed before the effects described by this schedule.
+    ///
+    /// - Parameter that: Schedule to be piped before this one.
+    public func reversePipe<State2, C>(_ that: Schedule<R, C, State2, A>) -> Schedule<R, C, (State2, State), B> {
```
miguelangel-dev: Maybe it is a dummy question, but I reverse like:
`public func reversePipe<State2, C>(_ that: Schedule<R, C, State2, B>) -> Schedule<R, C, (State2, State), A>`

truizlop: No, this is like the operators `<<<` and `>>>`, with the following signatures:

`>>>`: `(f: (A) -> B, g: (B) -> C) -> (A) -> C`
`>>>`: `(g: (B) -> C, f: (A) -> B) -> (A) -> C`

The only thing you are reversing is the order of the parameters, not their type arguments.

Sources/BowEffects/Data/Schedule.swift
```diff
@@ -0,0 +1,768 @@
...
+    /// - Parameter f: An effectful function that provides the delay time based on the output of the schedule.
+    public func addDelayM(_ f: @escaping (B) -> EnvIO<R, Never, TimeInterval>) -> Schedule<R, A, State, B> {
+        updated { upd in
+            { a, s in
+                let delay = EnvIO<R, Unit, TimeInterval>.var()
+                let s1 = EnvIO<R, Unit, State>.var()
+                return binding(
+                    delay <- f(self.extract(a, s)).unitError(),
+                    s1 <- upd(a, s),
+                    |<-EnvIO.sleep(delay.get),
```
miguelangel-dev: smart delay...

Sources/BowEffects/Data/Schedule.swift
miguelangel-dev: Ah! Now I see the point, thanks

Sources/BowEffects/Data/Internal/DispatchTimeInterval+Extensions.swift
miguelangel-dev: same here!

PR #480, opened 2019-12-02
Add scan methods to Traverse

ferranpujolcamins:
## Related issues

https://github.com/bow-swift/bow/issues/476

## Goal

Provide easy to use methods to perform stateful computations over a structure.

## Implementation details

Right now, the scan is not represented with a data structure like in [Haskell](http://hackage.haskell.org/package/foldl-1.4.5/docs/Control-Scanl.html#g:1). This means by now we cannot compose scans to only traverse once. But I believe this is an easy improvement that can be done in the future.

## Testing details

I just added some sanity tests to make sure the implementation makes sense. I don't know if other tests should be added, or even if it's worth merging the current ones.


Sources/Bow/Typeclasses/Traverse+Scan.swift
```diff
@@ -0,0 +1,131 @@
+import Foundation
+
+public extension Traverse {
```
truizlop: Move these extensions to the `Traverse.swift` file to keep them together and avoid too much segregation.

Tests/BowTests/Typeclasses/Traverse+Scan.swift
```diff
@@ -0,0 +1,82 @@
+import XCTest
+import Bow
+
+class TraverseScanTests: XCTestCase {
```
truizlop: I think these tests belong more to `ArrayK` than to `Traverse`, as they are particular to the `Traverse` instance of `ArrayK`, so I'd move them to `ArrayKTests.swift`.

In addition, it would be nice to think about properties of the `scanLeft` function to add them to `TraverseLaws.swift` so that they are tested for all instances of `Traverse`.

ferranpujolcamins: I think we could come up with some properties for scan, but those would follow from how it is actually implemented: traversing with the state monad. So I'm not sure it's worth adding property testing for it.

ferranpujolcamins:
I addressed the comments. Thank you.

Tests/BowTests/Typeclasses/Traverse+Scan.swift
truizlop: I think they are worth. If scan works for any `Traverse`, not only for arrays, the test should be verifiable for any `Traverse`, whereas these tests as biased towards the specific case of `ArrayK`.

ferranpujolcamins: I can think of the following test (scan with a pure function == map):
```swift
private static func pureFunctionScan() {
    property("Pure function scan") <~ forAll { (afa: KindOf<F, Int>, af: ArrowOf<Int, Int>) in
        let fa = afa.value
        let f = af.getArrow

        let sf: (Int) -> State<Void, Int> = { i in .init { ((), f(i)) } }
        return F.scanLeft(fa, initialState: (), f: sf) == F.map(fa, f)
    }
}
```
Also, from this it follows that scan with id == id.

However, the truth of this predicate follows from the definition of scan (I'm not sure the reasoning below is 100% rigorous, but I hope you can see my point): 
```swift
        F.scanLeft(fa, initialState: (), f: sf),
// Applying the definition of scan:
        F.traverse(fa, sf)^.runA(()),
// Let's us the uncurried version of traverse:
        F.utraverse(sf)(fa) |> runVoid,
        fa |> (runVoid <<< F.utraverse(sf)),
// And now by the naturality law of Traversal:
        (fa |> F.utraverse(Id.init <<< runVoid <<< sf))^.value,
        (F.utraverse(Id.init <<< runVoid <<< sf)(fa))^.value,
// Let's curry:
        F.traverse(fa, Id.init <<< runVoid <<< sf)^.value,
// sf ~= f
        F.traverse(fa, Id.init <<< f)^.value,
// Traverse a monadic Id function == map
        F.map(fa, f)
```

This is what I meant, if we think of scan as traversing with the state monad, the properties we can think will follow from the definition itself.

This is unlike the laws for tyeclasses. There, we have several implementations that we want to make sure work ok. But here, we only have one implementation, which happen to also be the very definition of what we want to implement.

truizlop:
Thanks for your contribution!

PR #484, opened 2019-12-04
Utilities for lens

truizlop:
## Goal

Combine two lenses with the same source but different focus.

### DIFF
--- a/Sources/BowOptics/DSL/At+Optics.swift
+++ b/Sources/BowOptics/DSL/At+Optics.swift
@@ -2,13 +2,13 @@ import Foundation
 import Bow
 
 // MARK: Extension when focus has `At` Instance
-public extension Lens where A: At {
+public extension Lens where A: At, S == T, A == B {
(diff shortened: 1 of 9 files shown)


Sources/BowOptics/DSL/At+Optics.swift
```diff
@@ -2,13 +2,13 @@ import Foundation
...
 // MARK: Extension when focus has `At` Instance
-public extension Lens where A: At {
+public extension Lens where A: At, S == T, A == B {
     /// Focuses on a specific index of this lens focus.
     ///
     /// - Parameter i: Index to focus.
     /// - Returns: A composed lens from this structure to the focused index.
     func at(_ i: A.AtIndex) -> Lens<S, A.AtFoci>  {
-        return self.fix + A.at(i)
+        return self + A.at(i)
```
miguelangel-dev: I think we could remove the `return` from this file

PR #493, opened 2020-01-05
[WIP] Add ZipArray

ferranpujolcamins:
## Related issues

I didn't open an issue because I thought it would be easier to discuss this looking at the whole code. Please feel free to close if you decide this does not belong in Bow.

This is experimental, I just aim for discussion, feedback, ideas...

Please let me know if anything is not clear enough.

## Goal

Have a ZipList-like type with applicative and traverse instances that never gets caught in an infinite loop.

The usual ZipList implementation relies on laziness and infinite lists. The problem with this approach is that it's tricky to tell whether a computation will terminate or not. In this PR I explore an approach that should allow us the query whether a list is infinite or not.

## Implementation details

`ZipArray` is a finite-lenght version of haskell's ZipList, where some operations might escape finiteness, but we can always check whether this has happened or not.

For applicative, ZipList (as implemented in Haskell) needs to implement `pure` by creating an infinite list with the argument. The trick here is to realize that the infinite list is used like an 'unknown length list': when you `ap` an infinite list with a finite list, you can then fix the length of the infinite list.

Thus, `ZipArray` has two cases: `.finite` and `.infinite`. The former indicates that we have a list of known length, the later indicates that we have a list of unknown length.

Furthermore, `ZipArray` can only be initialized with an array (i.e. a finite list). The only way you can create an infinite list is with `pure`. Thus, you cannot create a ZipList from an arbitrary swift infinite list (e.g. `sequence(first:next:)`). Thus, a `ZipArray` can always be expressed as a finite number of transformation of a `pure` *[insert proof here]*. This means a `ZipArray` is either finite, or a finite list followed by a single repeating element ad infinitum (e.g. `[a, b, c, d, d, d, d, ...]`)*[insert proof here]*.

Thus, the `.finite` case takes an array as parameter, the actual finite list of elements forming the `ZipArray`. The `.infinite` case takes a non-empty list as a paramater. This list represents the finite list where the last element is repeated ad-infinitum.

## What works
I could write implementations for `EquatableK`, `MonoidK`, `Functor` and `Applicative`.

## Problems
I don't know how to implement `Foldable`. I can fold the `.finite`, but not the `.infinite` case. Thus fold should have an `Option` return type, for the cases where we have an infinite list. Shall we create a new `MaybeFoldable`type class with such a signature? [1]

It'd be interesting to be able to implement `Traverse`, because then the transpose of a `matrix: ZipArray<ZipArray<A>>` can be expressed as `matrix.traverse(id)`.

## Alternatives
We could actually implement ZipList with lazy infinite lists. I actually did this [here](https://github.com/ferranpujolcamins/bow/tree/ziparray) using swift built-in LazySequences.

The drawbacks are that it's not nice to debug why your code is now stuck in an infinite loop. With the built-in lazy sequences, you need to remember to add `.lazy` all the time.

Maybe another option is to write a Bow type for infinite sequences base on `Eval` that offers better ergonomics, but if we want to be able to check for finiteness at runtime, we will most likely stumble with the same problems when trying to implement `Foldable` and `Traverse`.

---

[1] [I asked on stackoverflow about this](https://stackoverflow.com/questions/59603865/is-ther-a-type-class-for-fallible-fold)

We could add something like this:

```swift
public protocol FoldableT {
    associatedtype F

    static func foldLeft<A, B>(_ fa: Kind<Self, A>, _ b: B, _ f: @escaping (B, A) -> B) -> Kind<F, B>

    static func foldRight<A, B>(_ fa: Kind<Self, A>, _ b: Eval<B>, _ f: @escaping (A, Eval<B>) -> Eval<B>) -> Eval<Kind<F, B>>
}

public protocol Foldable: FoldableT where F == ForId {
    static func foldLeft<A, B>(_ fa: Kind<Self, A>, _ b: B, _ f: @escaping (B, A) -> B) -> B

    static func foldRight<A, B>(_ fa: Kind<Self, A>, _ b: Eval<B>, _ f: @escaping (A, Eval<B>) -> Eval<B>) -> Eval<B>
}

public extension Foldable {
    static func foldLeft<A, B>(_ fa: Kind<Self, A>, _ b: B, _ f: @escaping (B, A) -> B) -> Kind<F, B> {
        let r: B = foldLeft(fa, b, f)
        return Id(r)
    }

    static func foldRight<A, B>(_ fa: Kind<Self, A>, _ b: Eval<B>, _ f: @escaping (A, Eval<B>) -> Eval<B>) -> Eval<Kind<F, B>> {
        let r: Eval<B> = foldRight(fa, b, f)
        return r.map { Id($0) as Kind<ForId, B> }^
    }
}
```

Then:

```swift
extension ForZipArray: FoldableT {
     typealias F = ForOption

     static func foldLeft<A, B>(_ fa: Kind<Self, A>, _ b: B, _ f: @escaping (B, A) -> B) -> Kind<F, B> {
        switch fa^.data {
            case .finite(let array):
                return Option<B>.some(ForArrayK.foldLeft(ArrayK(array), b , f))
            case .infinite:
                return nil
        }
    }
}
```

`Foldable` adds the regular fold functions without the F embedding. With them, we can provide default implementations for the `FoldableT` functions that embed the result in `Id`. Thus, this new `Foldable` as a special case of `FoldableT` is a drop-in replacement of the normal `Foldable`, no code using it needs to be changed.

### DIFF
--- a/Bow.xcodeproj/project.pbxproj
+++ b/Bow.xcodeproj/project.pbxproj
@@ -329,6 +329,9 @@
 		8BA0F657217E2E9200969984 /* MonadCombine.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8BA0F4AD217E2E9200969984 /* MonadCombine.swift */; };
 		8BA0F65B217E2E9200969984 /* FunctorFilter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8BA0F4AE217E2E9200969984 /* FunctorFilter.swift */; };
 		8BB969AD2398661D005520D6 /* Queue.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8BB969AB239863F0005520D6 /* Queue.swift */; };
+		B53734D723C0800F00A012F3 /* ZipArray.swift in Sources */ = {isa = PBXBuildFile; fileRef = B53734D623C0800F00A012F3 /* ZipArray.swift */; };
+		B58DBFD323B964F000454CFA /* ZipArray+Gen.swift in Sources */ = {isa = PBXBuildFile; fileRef = B58DBFD223B964F000454CFA /* ZipArray+Gen.swift */; };
+		B58DBFD523B9654000454CFA /* ZipArrayTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = B58DBFD423B9654000454CFA /* ZipArrayTest.swift */; };
 		B8B910C0234D7F2600E44271 /* Semiring.swift in Sources */ = {isa = PBXBuildFile; fileRef = B8B910BF234D7F2600E44271 /* Semiring.swift */; };
 		B8B910C4234D846900E44271 /* SemiringLaws.swift in Sources */ = {isa = PBXBuildFile; fileRef = B8B910C3234D846900E44271 /* SemiringLaws.swift */; };
 		B8B910C6234DDA4000E44271 /* BoolInstancesTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = B8B910C5234DDA4000E44271 /* BoolInstancesTest.swift */; };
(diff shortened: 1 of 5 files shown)


truizlop:
Thanks for this contribution. I need to find time to read it carefully, but meanwhile, how do you think this fits with type classes [Semialign](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Semialign.kt) and [Align](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Align.kt)? We don't have issues for them yet, but I was planning to add them to our core.

ferranpujolcamins:
I'm not finding the time to investigate this further. Also it will probably be something belonging to a library, not Bow itself, since it's a rather unusual thing. I'm closing the PR. Apologies for the unresponsiveness from my side.

truizlop:
Don't worry. If you ever find time to contribute (this or anything else), you are always welcome!

PR #494, opened 2020-01-07
Add Comonad Transformers

truizlop:
## Goal

Add Comonad Transformers `EnvT`, `StoreT` and `TracedT`, as duals of the Monad Transformers `ReaderT`, `StateT` and `WriterT`, respectively.

## Implementation details

Implementation is based on the Haskell version of these transformers:

- [EnvT](https://hackage.haskell.org/package/comonad-5.0.6/docs/Control-Comonad-Trans-Env.html)
- [StoreT](https://hackage.haskell.org/package/comonad-5.0.6/docs/Control-Comonad-Trans-Store.html)
- [TracedT](https://hackage.haskell.org/package/comonad-5.0.6/docs/Control-Comonad-Trans-Traced.html)

For all these data types, a *plain* version, using `Id` as the inner Comonad, is provided: `Env`, `Store` and `Traced`.

Each of these data types provide instances for `ComonadEnv`, `ComonadStore` and `ComonadTraced`, respectively.

## Testing details

Laws for `Functor`, `Applicative` and `Comonad` are tested.


Sources/Bow/Transformers/EnvT.swift
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// MARK: Instance of `ComonadEnv` for `EnvT`
+
+extension EnvTPartial: ComonadEnv where W: Comonad {
+    public static func ask<A>(_ wa: Kind<EnvTPartial<E, W>, A>) -> E {
+        wa^.e
+    }
+}
+
```
miguelangel-dev: Should we provide an instance of ComonadTrans to EnvT?

```suggestion

// MARK: Instance of `ComonadTrans` for `EnvT`

extension EnvTPartial: ComonadTrans {
    public static func lower<W: Comonad, A>(_ twa: Kind<EnvTPartial, Kind<W, A>>) -> Kind<W, A> {
        twa.lower()
    }
}
```

PR #495, opened 2020-01-08
Day Convolution

### LINKED ISSUE #236: Port Day missing functions

truizlop:
## Related issues

- Closes #236 

## Goal

Provide correct instances for Day and add some missing functions.

## Implementation details

Implementation is based on a port from the [Haskell version](https://hackage.haskell.org/package/contravariant-0.6.1/docs/src/Data-Functor-Day.html). More info about Day can be found in [this blog post](https://blog.functorial.com/posts/2016-08-08-Comonad-And-Day-Convolution.html).

## Testing details

Laws for the instances are checked.

### DIFF
--- a/Sources/Bow/Data/Day.swift
+++ b/Sources/Bow/Data/Day.swift
@@ -4,29 +4,29 @@ public final class ForDay {}
 public final class DayPartial<F: Comonad, G: Comonad>: Kind2<ForDay, F, G> {}
 public typealias DayOf<F: Comonad, G: Comonad, A> = Kind<DayPartial<F, G>, A>
 
-public class Day<F: Comonad, G: Comonad, A> : DayOf<F, G, A> {
-    public static func fix(_ value : DayOf<F, G, A>) -> Day<F, G, A> {
-        return value as! Day<F, G, A>
+public final class Day<F: Comonad, G: Comonad, A> : DayOf<F, G, A> {
+    internal let left: Kind<F, Any /*B*/>
+    internal let right: Kind<G, Any /*C*/>
+    internal let f: (Any /*B*/, Any /*C*/) -> A
+
+    public init(left: Kind<F, Any>,
+                right: Kind<G, Any>,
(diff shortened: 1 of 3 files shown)


Sources/Bow/Data/Day.swift
```diff
@@ -4,29 +4,93 @@ public final class ForDay {}
...
 
-public class Day<F: Comonad, G: Comonad, A> : DayOf<F, G, A> {
-    public static func fix(_ value : DayOf<F, G, A>) -> Day<F, G, A> {
-        return value as! Day<F, G, A>
+public final class Day<F: Comonad, G: Comonad, A> : DayOf<F, G, A> {
+    internal let left: Kind<F, Any /*B*/>
+    internal let right: Kind<G, Any /*C*/>
+    internal let f: (Any /*B*/, Any /*C*/) -> A
+
+    public init(left: Kind<F, Any>,
```
miguelangel-dev: we could restrict unless the constructor to match the types, although we eare them internally.

example:
```swift
public init<B, C>(left: Kind<F, B>,
                right: Kind<G, C>,
                _ f: @escaping (B, C) -> A) {
        self.left = left.map { b in b as Any }
        self.right = right.map { c in c as Any }
        self.f = { (b: Any, c: Any) in f(b as! B, c as! C) }
    }
```

PR #496, opened 2020-01-08
Restore Co & Pairing

truizlop:
## Goal

Restore public visibility of `Co` and `Pairing` that was hidden in #483. Also, some cosmetic changes and finish the implementation of `Co`, which was not complete when it was introduced.


Sources/Bow/Data/Pairing.swift
```diff
@@ -1,17 +1,17 @@
...
+public class Pairing<F: Functor, G: Functor>: PairingOf<F, G> {
+    internal let pair: (Kind<F, (/*A*/Any) -> /*B*/Any>) -> (Kind<G, /*A*/Any>) -> /*B*/Any
     
-    static func fix(_ value : PairingOf<F, G>) -> Pairing<F, G> {
+    public static func fix(_ value : PairingOf<F, G>) -> Pairing<F, G> {
         value as! Pairing<F, G>
     }
     
-    init(_ pair: @escaping (Kind<F, (/*A*/Any) -> /*B*/Any>) -> (Kind<G, /*A*/Any>) -> /*B*/Any) {
+    public init(_ pair: @escaping (Kind<F, (/*A*/Any) -> /*B*/Any>) -> (Kind<G, /*A*/Any>) -> /*B*/Any) {
```
miguelangel-dev: we could add types to constructor like

```swift
public init<A, B>(_ pair: @escaping (Kind<F, (A) -> B>) -> (Kind<G, A>) -> B) {
        self.pair = { fab in pair(fab as! Kind<F, (A) -> B>) as! ((Kind<G, Any>) -> Any) }
}
```

miguelangel-dev: It'll require set `B` in `pair::Co`

```swift
public static func pair() -> Pairing<W, CoPartial<W>> {
        Pairing { (wab: Kind<W, (A) -> Any>) in { cowa in cowa^.run(wab) } }
}
```

Sources/Bow/Data/Co.swift
```diff
@@ -1,103 +1,86 @@
...
-            { cowa in
-                Co<W, /*A*/Any>.fix(cowa).runCo(wab)
-            }
-        }
+    public static func pair() -> Pairing<W, CoPartial<W>> {
+        Pairing { wab in { cowa in cowa^.run(wab) } }
     }
     
-    init(_ cow: @escaping /*forall R.*/(Kind<W, (A) -> /*R*/Any>) -> /*R*/Any) {
+    public init(_ cow: @escaping /*forall R.*/(Kind<W, (A) -> /*R*/Any>) -> /*R*/Any) {
```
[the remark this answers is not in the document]
truizlop: Fixing `R` makes the function concrete for that type, and not universal for all `R`. Therefore, we need to keep it under `Any` and fixing it in the `run` method.

PR #506, opened 2020-01-30
Bracket's guarantee does not work for errors

aballano:
As per [this issue in Arrow](https://github.com/arrow-kt/arrow/issues/1996) I just checked and it seems that Bow suffers from the same issue (unless I messed up with the tests 😅).

What changed:
- [x] Add test to BracketLaws to specifically check for guarantee working with an explicit error
- [ ] Add fix for Bracket, ideally fixing the above test

### DIFF
--- a/Tests/BowEffectsLaws/BracketLaws.swift
+++ b/Tests/BowEffectsLaws/BracketLaws.swift
@@ -94,7 +94,7 @@ public class BracketLaws<F: Bracket & EquatableK> where F.E: Equatable & Arbitra
     private static func bracketMustRunReleaseTask() {
         property("bracketMustRunReleaseTask") <~ forAll { (a: Int, e: F.E) in
             var msg = 0
-            return F.pure(a).bracket(release: { i in msg = i; return F.pure(()) }, use: { _ -> Kind<F, Int> in throw e })
+            return F.pure(a).bracket(release: { i in F.pure(()).map { msg = i } }, use: { _ -> Kind<F, Int> in throw e })
                 .attempt()
                 .map { _ in msg } == F.pure(a)
         }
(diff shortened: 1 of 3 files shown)


Tests/BowEffectsLaws/BracketLaws.swift
```diff
@@ -93,7 +94,17 @@ public class BracketLaws<F: Bracket & EquatableK> where F.E: Equatable & Arbitra
     private static func bracketMustRunReleaseTask() {
         property("bracketMustRunReleaseTask") <~ forAll { (a: Int, e: F.E) in
             var msg = 0
-            return F.pure(a).bracket(release: { i in msg = i; return F.pure(()) }, use: { _ -> Kind<F, Int> in throw e })
+            return F.pure(a).bracket(release: { i in F.pure(()).map { msg = i } }, use: { _ -> Kind<F, Int> in throw e })
```
miguelangel-dev: I think we could write it like: 
```suggestion
            return F.pure(a).bracket(release: { i in F.pure({ msg = i }) }, use: { _ -> Kind<F, Int> in throw e })
```

Tests/BowEffectsLaws/BracketLaws.swift
```diff
@@ -93,7 +94,17 @@ public class BracketLaws<F: Bracket & EquatableK> where F.E: Equatable & Arbitra
...
+                .attempt()
+                .map { _ in msg } == F.pure(a)
+        }
+    }
+    
+    private static func guaranteeMustRunFinalizerOnError() {
+        property("guaranteeMustRunReleaseOnError") <~ forAll { (a: Int, e: F.E) in
+            var msg = 0
+            let finalizer: Kind<F, ()> = F.pure(()).map { msg = a }
```
miguelangel-dev: same comment, here!

Tests/BowEffectsLaws/BracketLaws.swift
aballano: Almost! I ended with `F.pure(msg = a)`

PR #557, opened 2020-03-12
Semigroup and Monoid enhancements

truizlop:
## Related issues

The implementation of `combineAll` for Semigroup is not safe, as passing 0 values will make it crash. Semigroup always needs at least one value.

## Goal

- Force having at least one element in `combineAll` in the signature of the function.
- Provide an additional `combineAll` function in Monoid that can accept 0 or more values, using the `empty` value in case no values are combined.


miguelangel-dev:
Nice catch! Maybe for `Semigroup` will be useful to add helper methods to use `NEA`

truizlop:
As `combineAll` is meant to be use with variadic parameters, it is preferable not to add this API. If you have a `NEA` of Semigroup values, you can use `nea.fold()` to combine all of them.

PR #560, opened 2020-03-13
Composition series

truizlop:
## Goal

Composition is the essence of FP. This PR aims to provide documentation pages for different patterns of composition that are recurrent in multiple data types. It is captured as problems and solutions that people may have, together with examples. It covers the following cases:

- Composition of functions.
- Composition of Kleisli functions.
- Combination of values with Semigroup and Monoid.
- Transformation of values with Functor.
- Combination of independent computations with Applicative.
- Sequencing dependent computations with Monad.
- Transformation fo multiple effects with Traverse.

This section of the documentation may grow to include other type classes, but for this initial PR, it already contains quite a lot of content.


Documentation.app/Contents/MacOS/Composition.playground/Pages/Combining values of the same type.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,138 @@
...
+        ShoppingCart(
+            products: [],
+            amount: 0.0
+        )
+    }
+}
+/*:
+ ## Are all types that conform to Semigroup also conform to Monoid?
+ 
+ If your type conforms to `Monoid` it also must conform to `Semigroup`, but the reverse is not true. You may be able to combine values of a type, but not be able to provide an empty value.
```
miguelangel-dev: It will be nicer if we could embed this kind of paragraphs Q/A in "hidden" (collapsed) UI elements. - @calvellido ?

truizlop: Is there anything like that in standard Markdown? In this case, it is just this section, so I think it is not totally mandatory.

Documentation.app/Contents/MacOS/Composition.playground/Pages/Composing functions.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,103 @@
...
+// Provides the next talk for a Conference, if it exists.
+func nextTalk(at conference: Conference) -> Option<Talk>
+// nef:begin:hidden
+{ .none() }
+// nef:end
+/*:
+ Both functions are Kleisli functions, as their result are effectful, with Option being the effect. We can compose them into a function that provides the next Talk after a certain date, if it exists:
+ */
+let nextTalkAfterDate = Kleisli(nextConference(after:))
+    .andThen(Kleisli(nextTalk(at:)))
```
miguelangel-dev: Here, I think it helps if you add the type, and a little explanation about `Kleisli`


```suggestion
    .andThen(Kleisli(nextTalk(at:))) // Type: Kleisli<ForOption, Date, Talk> - represents a function with the signature (Date) -> Kind<ForOption, Talk>.
```

Documentation.app/Contents/MacOS/Composition.playground/Pages/Composing functions.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,103 @@
+// nef:begin:header
+/*
+ layout: docs
+ title: Composing functions
+ */
+// nef:end
+// nef:begin:hidden
+import Bow
```
miguelangel-dev: I think you need to import Foundation to use `Date`...

```suggestion
import Foundation
import Bow
```

miguelangel-dev:
I think it will be useful if we add the links to Bow API in each `typeclass`

Documentation.app/Contents/MacOS/Composition.playground/Pages/Composing functions.xcplaygroundpage/Contents.swift
truizlop: It seems it is not necessary.

PR #561, opened 2020-03-17
Add prisms auto-derivation

miguelangel-dev:
## Goal

Right now, in prisms, if the focus has associated values, we need to provide a bit of help by providing a pattern-matching function that extracts the associated values out of the case we are focusing on. 

Following the approach seen in [Point-Free](https://www.pointfree.co/episodes/ep89-case-paths-for-free) - I have updated it to auto derivate using reflection. So, no more boilerplate will be needed.

## Implementation details

Remove boilerplate in prisms using reflection for extracting the associated values from a `case`.

## Testing details

Added some tests for covering the next corner cases:
- Prism auto-derivation must support 2 cases: enums with/without associated types.
- Prism auto-derivation must support associated types with/without labelled.
- Prism auto-derivation must extract associated types in cases with the same name but with a different labelled parameter.


Tests/BowOpticsTests/PrismTest.swift
```diff
@@ -122,4 +122,42 @@ class PrismTest: XCTestCase {
...
+    }
+    
+    enum EnumLabeled: AutoPrism {
+        case labeled(label: String)
+        case labeled(anotherLabel: String)
+    }
+    
+    func testAutoDerivatePrism_WithoutAssociatedTypes() {
+        let prism = Authentication.prism(for: Authentication.unkown)
+        XCTAssert(prism.getOption(.unkown).isDefined)
```
truizlop:
```suggestion
        XCTAssertTrue(prism.getOption(.unkown).isDefined)
```

Tests/BowOpticsTests/PrismTest.swift
```diff
@@ -122,4 +122,42 @@ class PrismTest: XCTestCase {
...
+    }
+    
+    func testAutoDerivatePrism_WithAssociatedTypes() {
+        let prism = Authentication.prism(for: Authentication.authorized)
+        let result = prism.getOption(.authorized(8, "information")).toOptional() ?? (0, "")
+        XCTAssert(result == (8, "information"))
```
truizlop:
```suggestion
        XCTAssertTrue(result == (8, "information"))
```

Tests/BowOpticsTests/PrismTest.swift
```diff
@@ -122,4 +122,42 @@ class PrismTest: XCTestCase {
...
+    }
+    
+    func testAutoDerivatePrism_WithLabeledAssociatedTypes() {
+        let prism = Authentication.prism(for: Authentication.requested)
+        let result = prism.getOption(.requested(1, info: "information")).toOptional() ?? (0, "")
+        XCTAssert(result == (1, "information"))
```
truizlop:
```suggestion
        XCTAssertTrue(result == (1, "information"))
```

Sources/BowOptics/Prism.swift
```diff
@@ -421,19 +421,6 @@ public extension Prism where S == A, S == T, A == B {
...
-    /// - Parameters:
-    ///   - extract: Function to extract a value from the source.
-    ///   - embed: Function to embed a value into the source.
-    convenience init(extract: @escaping (S) -> A?, embed: @escaping (A) -> S) {
-        self.init(
-            getOrModify: { s in extract(s).flatMap(Either.right) ?? .left(s) },
-            reverseGet: embed)
-    }
-}
-
```
truizlop: This constructor is still useful for cases where you want to build the Prism yourself. Please, revert this change.

miguelangel-dev: Sure! Make sense :)

PR #565, opened 2020-03-23
Bring consistency across HKT with partial application

truizlop:
## Goal

Currently, there is some dissonance between HKTs with one type argument and others with 2 or more arguments. Whereas the latter have their `Partial` versions, the former do not have this alias. This results in the following when instances for type classes need to be provided:

```swift
extension ForOption: Functor { ... }
extension EitherPartial: Functor { ... }
```

It seems weird not to have a single convention when implementing instances of type classes on the kind level, and having to remember that, for types with one type argument you need to use the `For` form, and for others the `Partial` form.

This PR unifies the convention by providing a `Partial` alias for types with a single type argument, so that the previous code still works, but the preferred way of doing it is:

```swift
extension OptionPartial: Functor { ... }
extension EitherPartial: Functor { ... }
```

This way, every time one needs to provide an instance for a type class in the kind level, only needs to remember that it is done in the `Partial` type.

Besides this, the PR updates the code of the modified files to make aesthetic changes:

- Remove returns when possible.
- Use `TOf<A>` instead of `Kind<T, A>` when possible.
- Use `^` instead of `fix` when possible.
- Break down lines and simplify some code for better readability.

Documentation has been updated accordingly.

### DIFF
--- a/Sources/Bow/Data/Id.swift
+++ b/Sources/Bow/Data/Id.swift
@@ -111,7 +111,7 @@ extension IdPartial: Bimonad {}
 
 // MARK: Instance of `Foldable` for `Id`
 extension IdPartial: Foldable {
-    public static func foldLeft<A, B>(_ fa: Kind<ForId, A>, _ b: B, _ f: @escaping (B, A) -> B) -> B {
+    public static func foldLeft<A, B>(_ fa: IdOf<A>, _ b: B, _ f: @escaping (B, A) -> B) -> B {
         f(b, fa^.value)
     }
 
--- a/Sources/Bow/Data/Id.swift
+++ b/Sources/Bow/Data/Id.swift
@@ -3,6 +3,9 @@ import Foundation
 /// Witness for the `Id<A>` data type. To be used in simulated Higher Kinded Types.
 public final class ForId {}
 
+/// Partial application of the Id type constructor, omitting the last type parameter.
+public typealias IdPartial = ForId
+
 /// Higher Kinded Type alias to improve readability of `Kind<ForId, A>`.
 public typealias IdOf<A> = Kind<ForId, A>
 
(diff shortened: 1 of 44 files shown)


Sources/Bow/Data/Option.swift
```diff
@@ -198,108 +198,104 @@ extension ForOption: Monad {
...
+                                    { y in .some(x.combine(y)) })
         })
     }
 }
 
 // MARK: Instance of `Semigroupal` for `Option`,
-extension ForOption: Semigroupal {
-    public static func product<A, B>(_ a: Kind<ForOption, A>, _ b: Kind<ForOption, B>) -> Kind<ForOption, (A, B)> {
+extension OptionPartial: Semigroupal {
+    public static func product<A, B>(_ a: OptionOf<A>, _ b: Kind<ForOption, B>) -> OptionOf<(A, B)> {
```
miguelangel-dev:
```suggestion
    public static func product<A, B>(_ a: OptionOf<A>, _ b: OptionOf<B>) -> OptionOf<(A, B)> {
```

Sources/Bow/Data/Option.swift
```diff
@@ -69,58 +72,58 @@ public final class Option<A>: OptionOf<A> {
     /// - Parameter predicate: Boolean predicate to test the wrapped value.
     /// - Returns: This value if it does not match the predicate, or none otherwise.
     public func filterNot(_ predicate: @escaping (A) -> Bool) -> Kind<ForOption, A> {
```
miguelangel-dev:
```suggestion
    public func filterNot(_ predicate: @escaping (A) -> Bool) -> OptionOf<A> {
```

Sources/Bow/Data/Id.swift
```diff
@@ -94,50 +96,47 @@ extension ForId: Monad {
...
 }
 
 // MARK: Instance of `Bimonad` for `Id`
-extension ForId: Bimonad {}
+extension IdPartial: Bimonad {}
 
 // MARK: Instance of `Foldable` for `Id`
-extension ForId: Foldable {
+extension IdPartial: Foldable {
     public static func foldLeft<A, B>(_ fa: Kind<ForId, A>, _ b: B, _ f: @escaping (B, A) -> B) -> B {
```
miguelangel-dev:
```suggestion
    public static func foldLeft<A, B>(_ fa: IdOf<A>, _ b: B, _ f: @escaping (B, A) -> B) -> B {
```

PR #571, opened 2020-03-25
Clean Function1 tests

miguelangel-dev:
patch to [PR](https://github.com/bow-swift/bow/pull/570)


truizlop:
Test module will be addressed later, after another refactor is done.

PR #576, opened 2020-03-26
Cleanup Typeclasses folder in Core module

truizlop:
## Goal

Cleanup code in the Typeclasses folder in the Core module by:

- Using `FOf<A>` instead of `Kind<F, A>`.
- Removing backticks in `MARK` sections as they do not render properly in Jazzy.
- Removing returns.
- Replacing `fix` by `^`.

### DIFF
--- a/Sources/Bow/Typeclasses/Traverse.swift
+++ b/Sources/Bow/Typeclasses/Traverse.swift
@@ -8,7 +8,9 @@ public protocol Traverse: Functor, Foldable {
     ///   - fa: A structure of values.
     ///   - f: A function producing an effect.
     /// - Returns: Results collected under the context of the effect provided by the function.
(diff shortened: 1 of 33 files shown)


Sources/Bow/Typeclasses/Traverse.swift
```diff
@@ -89,21 +95,21 @@ public extension Traverse {
...
 
         let stateStep: (A) -> State<B, B> = { a in
             let previousValue = State<B, B>.var()
             let nextValue = State<B, B>.var()
 
             return binding(
-                previousValue <- StateTPartial<ForId, B>.get(),
-                nextValue     <- StateTPartial<ForId, B>.pure(f(previousValue.get, a)),
-                              |<-StateTPartial<ForId, B>.set(nextValue.get),
+                previousValue <- .get(),
```
miguelangel-dev: nice!

PR #577, opened 2020-03-26
Cleanup Data folder in Effects module

truizlop:
## Goal

Cleanup code in the Data folder in the Effects module by:

- Using `FOf<A>` instead of `Kind<F, A>`.
- Removing backticks in `MARK` sections as they do not render properly in Jazzy.
- Removing returns.
- Replacing `fix` by `^`.

### DIFF
--- a/Sources/BowEffects/Data/IO.swift
+++ b/Sources/BowEffects/Data/IO.swift
@@ -98,7 +98,9 @@ public class IO<E: Error, A>: IOOf<E, A> {
     ///   - fe: Function to transform the error type argument.
     ///   - fa: Function to transform the output type argument.
     /// - Returns: An IO with both type arguments transformed.
-    func bimap<EE: Error, B>(_ fe: @escaping (E) -> EE, _ fa: @escaping (A) -> B) -> IO<EE, B> {
+    func bimap<EE: Error, B>(
+        _ fe: @escaping (E) -> EE,
+        _ fa: @escaping (A) -> B) -> IO<EE, B> {
         mapError(fe).map(fa)^
     }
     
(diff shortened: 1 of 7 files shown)


Sources/BowEffects/Data/IO.swift
```diff
@@ -404,8 +420,12 @@ public class IO<E: Error, A>: IOOf<E, A> {
     ///   - policy: Repeating policy.
     ///   - onUpdateError: A function providing an error in case the policy fails to update properly.
     /// - Returns: A computation that is repeated based on the provided policy when it succeeds.
-    public func `repeat`<S, O>(_ policy: Schedule<Any, A, S, O>, onUpdateError: @escaping () -> E) -> IO<E, O> {
-        self.env().repeat(policy, onUpdateError: onUpdateError).provide(())
+    public func `repeat`<S, O>(
+        _ policy: Schedule<Any, A, S, O>,
+        onUpdateError: @escaping () -> E = { fatalError("Impossible to update error on repeat.") }) -> IO<E, O> {
```
miguelangel-dev: not sure about it... which is the goal about it?

miguelangel-dev: I see it as an optional parameter, but in the end, it is necessary...so, why to give a default-impossible state?

truizlop: This function should not be necessary in most cases, as there won't be an error that is not of the correct type. However, the way it is implemented, forces to go from `Error` to `E` and if you want to cover all possible cases, you'd need the `onUpdateError` function. In practice, as I said, it should never happen, and when I have had to use it, I've found myself having to pass the `fatalError` to make the compiler happy, so this is just a convenience to avoid having to do it every single time.

miguelangel-dev: Ok, thanks for clarifying it.

PR #584, opened 2020-04-03
Remove transitive test-only dependencies

truizlop:
Currently, Bow depends on SwiftCheck to write tests and exposes some libraries for testing laws. SwiftCheck exposes dependencies that are only used for testing, but end up being fetched every time Bow is imported. One of this dependencies causes SwiftUI previews not to load at all because it encounters compilation errors when compiling this dependency.

Given the transitive dependencies for SwiftCheck are only used for testing, they should be removed. Therefore, as Swift Package Manager still does not allow to specify a dependency is only for testing, we have created a fork of SwiftCheck, where the `Package.swift` is modified to remove the test targets and the dependencies. The `Package.swift` in Bow now depends on this fork. Consumers of Bow will not fetch the transitive dependencies anymore, thus fixing the SwiftUI preview problem.

### DIFF
--- a/Bow.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Bow.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -2,21 +2,21 @@
   "object": {
     "pins": [
       {
+        "package": "Chalk",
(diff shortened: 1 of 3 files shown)


miguelangel-dev:
Good workaround!

Issue #603, opened 2020-07-10
["Request"] Add extract and embed properties to Prism


ferranpujolcamins:
## Description

Bow offers the possibility to create a Prism providing an `extract` and an `embed` functions.
For consistency, Prism should allow such functions to be retrieved from a Prism.

It might help people who is not used to the Either formulation of Prisms.

## Potential implementation
```swift
public extension Prism where S == T, A == B {
    var extract: (S) -> A? {
        getOrModify >>> \.orNil
    }

    var embed: (A) -> S {
        reverseGet
    }
}
```

## Modules

Optics

truizlop:
LGTM, please open a PR with that. There is a way of creating a Prism with embed and extract, but not a way of retrieving these properties.

PR #604, opened 2020-07-13
Add extract and embed properties to Prism

ferranpujolcamins:
## Related issues

#603 

## Goal

There is a way of creating a Prism with embed and extract, but not a way of retrieving these properties.


truizlop:
Thanks!

PR #610, opened 2020-09-21
Conform Option to ExpressibleByNilLiteral protocol

dsabanin:
Adds an ability to use nil values where `Option<A>` values are expected, converting them to `Option<A>.none()` automatically. 

Works in assignments, function call arguments, function default argument values, etc.

#### Nil assignment to `Option<A>`
```swift
let a: Option<A> = nil 
// is equivalent to
let a: Option<A> = Option.none()
```

#### Function call argument values
```swift
func test(a: Option<A>, b: Option<B> = nil) { ... }
test(nil) 
// is equivalent to
func test(a: Option<A>, b: Option<B> = Option.none()) { ... }
test(Option.none()) 
```

## Goal

Make Bow's `Option` type feel more native to Swift, similar to how `Optional` behaves in Bow Lite.

## Implementation details

Surprisingly simple conformance to `ExpressibleByNilLiteral` was enough to get both the nil assignment to act as `Option<A>.none()` assignment *and* default initialization values for `Option<A>` variables. I did have to change access control from private to `fileprivate` on `Option` class in order to have `ExpressibleByNilLiteral` conformance done as an extension to fit the existing code styles.

## Testing details

I was unable to compile the test suite (without my changes), due to compilation time outs in BowLaws code in the `ApplicativeLaws` class. I did however write the tests and made sure they work on a different codebase.

### DIFF
--- a/Sources/Bow/Data/Option.swift
+++ b/Sources/Bow/Data/Option.swift
@@ -39,7 +39,7 @@ public final class Option<A>: OptionOf<A> {
         Option(a)
     }
 
-    private init(_ value: A?) {
+    fileprivate init(_ value: A?) {
         self.value = value
     }
 
@@ -429,4 +429,10 @@ extension BidirectionalCollection {
     public func singleOrNone(_ predicate: (Element) -> Bool) -> Option<Element> {
         self.filter(predicate).singleOrNone
     }
+}
+
+extension Option: ExpressibleByNilLiteral {
+   public convenience init(nilLiteral: ()) {
+      self.init(nil)
+   }
 }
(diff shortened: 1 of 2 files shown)


truizlop:
Fantastic! I'll check the issues you mention about Applicative Laws, but it seems CI was able to run the tests nicely. Please, let me know your Xcode version to try to reproduce the issue. Thanks for your contribution!

truizlop (approved):

dsabanin:
> Fantastic! I'll check the issues you mention about Applicative Laws, but it seems CI was able to run the tests nicely. Please, let me know your Xcode version to try to reproduce the issue. Thanks for your contribution!

That's awesome, thank you! Good to know about the test suite, and you're right, I keep forgetting that living on the edge has a price – I'm on **Xcode 12.2 beta (12B5018i)**, on **Big Sur 11.0 Beta (20A5374g)**. I think it makes sense to ignore this as a fluke if CI doesn't have this issue, I just didn't want to assume anything.

truizlop:
We'll keep an eye in future versions of Xcode, I know that sometimes it causes compilation failures due to timeouts in the type inference. Anyways, thanks for your first contribution!

Issue #611, opened 2020-09-22
["Request"] Improve binding with function builders


ferranpujolcamins:
## Description

By using a very simple function builder we can make the binding function a bit easier to use by not requiring commas after each binding.

## Sample usage

```swift
func product(_ a: ArrayK<Int>, _ b: ArrayK<Int>) -> Kind<ForArrayK, String> {
    let x = ArrayK<Int>.var()
    let y = ArrayK<Int>.var()
    return binding({
        x <- a // Note that there are no commas here.
        y <- b
    }, yield: "(\(x), \(y))")
}

```

## Potential implementation

```swift
public func binding<F: Monad, A>(
    @BindingsBuilder _ instructions: () -> [BindingExpression<F>],
    yield value: @autoclosure @escaping () -> A) -> Kind<F, A> {
    MonadComprehension<F>.buildBlock(instructions(), yield: value)
}

@_functionBuilder
struct BindingsBuilder {
    static func buildBlock<F>(_ instructions: BindingExpression<F>...) -> [BindingExpression<F>] {
        instructions
    }
}
```

## Modules

Bow

## Breaking changes

No breaking changes, this new binding function can coexist with the existing one.

truizlop:
I think you've read my mind! I was thinking about adding this a couple of days ago. The only change I'd make from your proposal is using a closure for `yield` as well, in order to leverage the multiple trailing closures evolution proposal, being the new syntax something like:

```swift
func product(_ a: ArrayK<Int>, _ b: ArrayK<Int>) -> Kind<ForArrayK, String> {
    let x = ArrayK<Int>.var()
    let y = ArrayK<Int>.var()
    return binding {
        x <- a // Note that there are no commas here.
        y <- b
    } yield: {
        "(\(x), \(y))"
    }
}
```

What do you think?

ferranpujolcamins:
Awesome! Do you already have something prepared? Or may I do the honours?

truizlop:
Please, go ahead!

PR #614, opened 2020-09-23
Add bindings overload based on FunctionBuilders

ferranpujolcamins:
## Related issues

#611 

Requires Xcode 12


truizlop:
CI is failing because of the need of Xcode 12, but as far as I can tell, this is only due to the multiple trailing closures in the tests, right? We can write this in Swift 5.2 style to avoid needing to push CI to use Xcode 12, as we still need to adapt nef to work with this version and we won't be able to check the docs.

truizlop (approved):

truizlop:
Thanks for the contribution!

PR #619, opened 2020-09-30
Revamp Coyoneda

### LINKED ISSUE #295: Port Coyoneda missing tests
## Description

In order to align Bow and [Arrow](https://github.com/arrow-kt/arrow) API, there are some missing tests that require to be ported

* Check [Coyoneda](https://github.com/arrow-kt/arrow/blob/master/modules/free/arrow-free-data/src/test/kotlin/arrow/free/CoyonedaTest.kt) tests in Arrow repository
* Port the missing tests

truizlop:
## Related issues

Close #295 

## Goal

Revamp the implementation for Coyoneda and complete its API.


Sources/BowFree/Coyoneda.swift
```diff
@@ -2,66 +2,157 @@ import Foundation
...
+    public static func fix(_ fa: CoyonedaOf<F, A>) -> Coyoneda<F, A> {
+        fa as! Coyoneda<F, A>
     }
+}
 
-    public init(_ pivot : Kind<F, P>, _ ks : [AnyFunc]) {
-        self.pivot = pivot
-        self.ks = ks
+public extension Coyoneda where F: Functor {
+    static func liftCoyoneda(_ fa: Kind<F, A>) -> Coyoneda<F, A> {
```
ferranpujolcamins, line 26: In Haskell [`liftCoyoneda`](https://hackage.haskell.org/package/kan-extensions-5.2/docs/Data-Functor-Coyoneda.html#v:liftCoyoneda) does not require F to be a functor. Can we achieve this here too?

ferranpujolcamins: I'm trying to port Haskell's Program (Free-er monad), and I need this here: http://hackage.haskell.org/package/reasonable-operational-0.1.0.1/docs/src/Control-Monad-Operational.html#singleton

ferranpujolcamins:
Can Coyoneda have an Arbitrary instance?

Sources/BowFree/Coyoneda.swift
```diff
@@ -2,66 +2,157 @@ import Foundation
...
-    fileprivate let pivot: Kind<F, P>
-    fileprivate let ks: [AnyFunc]
-
-    public static func apply(_ fp : Kind<F, P>, _ f : @escaping (P) -> A) -> Coyoneda<F, P, A> {
-        return unsafeApply(fp, [f as! AnyFunc])
-    }
-
-    public static func unsafeApply(_ fp : Kind<F, P>, _ fs : [AnyFunc]) -> Coyoneda<F, P, A> {
-        return Coyoneda<F, P, A>(fp, fs)
+public class Coyoneda<F, A>: CoyonedaOf<F, A> {
```
ferranpujolcamins, line 8: Can't Coyoneda be final?

truizlop: Yes, I forgot to add the final keyword. The Arbitrary instance is also in place now.

Sources/BowFree/Coyoneda.swift
truizlop: I've not been able to remove this. The problem is that I cannot go from `Kind<F, A>` to `Kind<F, Any>` without `map`ing to do the cast. Any idea how we can achieve this?

ferranpujolcamins:
Looks good to me. Thank you!

miguelangel-dev (approved):

PR #621, opened 2020-10-01
Implement Coyoneda based on Exists

ferranpujolcamins:
## Goal

Implement Coyoneda in a way that avoids the Functor requirement on F for `liftCoyoneda`[1].

### Exists
There's a new type called `Exists` that encodes the notion of existential type. `Exists<F>` is a wrapper around a `Kind<F, X>` value, where X is some type that is not exposed on `Exists` signature. You provide a value of type `Kind<F, X>` when constructing an `Exists<F>` and the specific type X that you provide is hidden. 

The only way to interact with the contents of `Exists<F>` is with a function of type `Kind<F, X> -> R` that is polymorphic on X (represented by the new class `CokleisliK<F, R>`). The idea is that you pass a `CokleisliK<F, R>` to an `Exists<F>` and the function is called with the specific type inside the Exists. This is done by the Exists' method `func run<R>(_ f: CokleisliK<F, R>) -> R`.

### Coyoneda
Coyoneda is now `Exists<CoyonedaFPartial<F, A>>`. `CoyonedaF<F, A, P>` holds the `pivot: Kind<F, P>` and the function `f: (P) -> A`. `Exists<CoyonedaFPartial<F, A>>` effectively hides the specific type P that resides inside the coyoneda instance.

## Pros
There's not a single force cast or ugly hacks inside Coyoneda. Any trick we need to do will be located only inside the implementation of Exists' `run`. 
Furthermore, `liftCoyoneda` does no longer require F to be a functor [1].

`Exists` can be used for other things other than Coyoneda. We can see it used in [Day convolution](https://github.com/paf31/purescript-day/blob/23a4aa9c77854edc4ff28258a113dee7a9d600f5/src/Data/Functor/Day.purs#L39) and in the wild:
- [1](https://github.com/rohanpanda001/fruit-basket/blob/311f0fb4df9318828c7e3ab41a6d21cbf460ebbb/bower_components/purescript-profunctor/src/Data/Profunctor/Split.purs#L16)
- [2](https://github.com/TinkerTravel/purescript-config/blob/f3031d8ba5d0ae61c84b91a70f3e3f04ee2a4b08/src/Data/Config.purs#L25)
- [3](https://github.com/boygao1992/ADTplayground/blob/681c2833a9df612f67ceb91df89d7512c8bb0a47/purescript_free_cofree/src/NaiveCoroutine.purs#L45)
- [4](https://github.com/b123400/purescript-pux-form/blob/6491bb4e74d2b016abe62672ed5748d150187412/src/Pux/Form.purs#L26)

## Cons
Boilerplate: Because Exists' `run` needs a polymorphic function, lots of functions need to be wrapped inside a class.

## Problems
**I don't know how to implement Exists' `run`function**. Essentially, what we want is to inspect the runtime type of an `Any` value `a` and call a generic function binding the generic parameter to the runtime type of `a`. If you just pass the `Any` value to the generic function, the generic parameter is bound to `Any`. This is not what we want.

You can inspect the runtime type of an `Any` variable and cast it using `case let any as String` in a switch statement. However, you need to manually add each type to the switch, which means you cannot define `run` in a truly generic way. 
Indeed this is how `run` is naively (and incompletely) implemented now: I manually added a case for each type that was passed to `run` during the tests. The tests pass, which means that if we can implement `run` fully polymorphic, everything should work as described here.

Because of the temporary funky implementation for Exists' `run`, I needed to move Exists to the BowFree module, but if we find a solution to the implementation problem we will be able to move Exists to Bow.

## Notes
[1] We need `liftCoyoneda` if we want to have a shot at porting haskell's `Program` type for free-er monads. Specifically we need this for the [singleton constructor](https://github.com/bow-swift/bow/pull/hackage.haskell.org/package/free-operational-0.5.0.0/docs/src/Control-Monad-Operational-Simple.html)

## References
Exists and the Coyoneda implementation based on it are ported from Purescript:
- [purescript-exists](https://github.com/purescript/purescript-exists) 
- [coyoneda](https://github.com/purescript/purescript-free/blob/v4.0.0/src/Data/Coyoneda.purs#L30)

## Testing details

To test `Exists` I ported the [purescript tests](https://github.com/purescript/purescript-exists/blob/master/test/Test/Main.purs) which consist of a Stream datatype based on existentials and some operations on it.

`Coyoneda` tests all pass with the new implementation.


ferranpujolcamins:
Do you have any idea how Exists' `run` could be properly implemented?
Otherwise I'll ask the swift forums to know if this is even posible.

PR #623, opened 2020-10-04
Implement Coyoneda based on Exists

ferranpujolcamins:
Successor of #621, please refer to the description of that Pr for details.

I was able to find a way to implement `Exists` `run` method without any ugly hacks or force casts.

Coyoneda Tests all pass. I added a simple property-based test for Exists, I cannot think of any additional properties nor have I found any references. In purescript, they only have the test based on the implementation of Stream.

The implementation of Free-er monad (Haskell's Program) as `Free<Coyoneda<F>>` (coming soon) works with this PR 🎉 

[Here](https://github.com/ferranpujolcamins/bow/pull/1) I have additional changes for this PR that make Coyoneda's map function stack safe. I kept this changes separated from the main PR for now to keep things easier to digest.


truizlop (approved):
This is a very clever implementation, thank you so much for adding it here! I am wondering if we could leverage this to improve the castings we do in `Day` and `Pairing`...

truizlop:
@ferranpujolcamins can you take a look at the conflicts before we merge this?

ferranpujolcamins:
Conflicts fixed!

PR #624, opened 2020-10-06
New tailRecM laws + Bug fixes 🐞

ferranpujolcamins:
I added two new laws to test `tailRecM` (I took them [from Scala](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/FlatMapLaws.scala#L33)).

With the new `tailRecMConsistentFlatMap` law I discovered and fixed bugs in the implementation of `tailRecM` for `Ior` and `WriterT`.

- In `Ior`, when flatMapping on a `Both` value, when the new value is `Right`, we were returning `Both` (correct value is `Right`).
- In `WriterT` we were not combining the old `W` with the new `W` obtained after flatmapping.


truizlop (approved):

truizlop:
Awesome! Good catch!

PR #630, opened 2020-10-14
Make Coyoneda map stack safe

ferranpujolcamins:
Currently Coyoneda uses function composition to join the inner function with a function we map over. The problem with this is that if you try to evaluate[1] a coyoneda value that has been mapped a lot of times you get a stack overflow.

This PR solves this issue by using a new `Function1LazyComposition` that keeps a type-erase list of functions that can be called one after the other.

This solution is similar to how Coyoneda was implemented in Bow before, and also how it is implemented in [Scala](https://github.com/typelevel/cats/blob/master/free/src/main/scala/cats/free/Coyoneda.scala), but `Function1LazyComposition` keeps track of what the input and output types of the composed function are.

[1] For example when calling `lower()` or `flatMap`


truizlop (commented):
The implementation looks right but I'd choose a different name for the type. What about `LazyFunction1`?

ferranpujolcamins:
Done!

Sources/Bow/Arrow/LazyFunction1.swift
```diff
@@ -0,0 +1,57 @@
...
+
+fileprivate func erase<I, O>(_ f: @escaping (I) -> O) -> (Any) -> Any {
+    { f($0 as! I) }
+}
+
+public postfix func ^<I, O>(_ fa: LazyFunction1Of<I, O>) -> LazyFunction1<I, O> {
+    LazyFunction1.fix(fa)
+}
+
+// MARK: Instance of Functor for Function1LazyComposition
```
truizlop, line 50:
```suggestion
// MARK: Instance of Functor for LazyFunction1
```

Tests/BowGenerators/Arrow/LazyFunction1+Gen.swift
```diff
@@ -0,0 +1,20 @@
...
+
+extension LazyFunction1: Arbitrary where I: CoArbitrary & Hashable, O: Arbitrary {
+    public static var arbitrary: Gen<LazyFunction1<I, O>> {
+        ArrowOf<I, O>.arbitrary.map { arrow in
+            LazyFunction1(arrow.getArrow)
+        }
+    }
+}
+
+// MARK: Instance of ArbitraryK for Function1
```
truizlop, line 14:
```suggestion
// MARK: Instance of ArbitraryK for LazyFunction1
```

Tests/BowTests/Arrow/LazyFunction1Test.swift
```diff
@@ -0,0 +1,40 @@
+import XCTest
+import SwiftCheck
+@testable import BowLaws
+import Bow
+
+extension LazyFunction1Partial: EquatableK where I == Int {
+    public static func eq<A>(_ lhs: Kind<LazyFunction1Partial<I>, A>, _ rhs: Kind<LazyFunction1Partial<I>, A>) -> Bool where A : Equatable {
```
truizlop, line 7:
```suggestion
    public static func eq<A: Equatable>(_ lhs: LazyFunction1Of<I, A>, _ rhs: LazyFunction1Of<I, A>) -> Bool {
```

truizlop (changes requested):
Please, add documentation for the `LazyFunction1` methods.

ferranpujolcamins:
Done

ferranpujolcamins:
Should be ready now

truizlop (approved):

truizlop:
Awesome!

PR #634, opened 2020-10-16
Introduce Program type for free-er monads

ferranpujolcamins:
## Goal

(Re)Introduce[1] Free-er monads in Bow. `Program` allows you to get a monad from any type constructor. Unlike `Free`, it doesn't require `F` to be a functor, so not only does it defer the evaluation of `flatMap` calls for later interpretation into another monad, but also (unlike Free) defers calls to `map`.

Reference: [free-operational](http://hackage.haskell.org/package/free-operational-0.5.0.0/docs/Control-Monad-Operational-Simple.html)

**Do we need both `Free` and `Program`?**
Yes. When you use `Free`, certain properties can be enforced in the Functor implementation of `F`, meaning that those properties will be fulfilled no matter what interpreter you use later [2].

When you use `Program`, you cannot enforce properties on the Functor implementation of F, because it's just not a functor. But, on the other hand, you don't have to implement the Functor instance. So `Program` is easier to use, but you cannot guarantee custom properties hold. Context will tell what option is best for you.

## Implementation details

`Program` is just `Free<Coyoneda<F>>`. `Free` needs a functor, since `F` is not a functor, we use `Coyoneda<F>` which is the free functor of `F` and defers evaluation of `map`. So, at the end, `Program` defers the evaluation of `flatMap` calls for later interpretation into another monad, but unlike `Free`, it also defers calls to `map` (thanks to the inner `Coyoneda`).

## Testing details

I used the same test program found in the `Free` tests, but you can see how it is not a functor here!

## Name
I'm not really happy with the name *Program*. I took it from Haskell, but it doesn't really tell what this type is about IMHO. 

On the other hand, *Freer* is too similar to `Free` and I think this can cause confusion or problems in code.

A third option would be to come up with a new name that tries to capture the essence of this data type but is not similar to `Free`. Something like `EvenFreer`? But then it would be more difficult to relate this type to its Haskell and Scala counterpart.

What's your view on the naming issue?

## Notes
[1] The previous interpretation of Free was actually the Free-er monad! So we are just bringing it back.
[2] See [Purify code using free monads (*Proofs* section)](http://www.haskellforall.com/2012/07/purify-code-using-free-monads.html) for an example.

### DIFF
--- a/Sources/BowFree/Program.swift
+++ b/Sources/BowFree/Program.swift
@@ -66,27 +66,27 @@ public extension Program where F: Monad {
 }
 
 extension ProgramPartial: Functor {
-    public static func map<A, B>(_ fa: Kind<ProgramPartial<F>, A>, _ f: @escaping (A) -> B) -> Kind<ProgramPartial<F>, B> {
+    public static func map<A, B>(_ fa: ProgramOf<F, A>, _ f: @escaping (A) -> B) -> ProgramOf<F, B> {
         Program(asFree: fa^.asFree.map(f)^)
     }
 }
 
 extension ProgramPartial: Applicative {
-    public static func pure<A>(_ a: A) -> Kind<ProgramPartial<F>, A> {
+    public static func pure<A>(_ a: A) -> ProgramOf<F, A> {
         Program(asFree: Free<CoyonedaPartial<F>, A>.pure(a)^)
     }
 
-    public static func ap<A, B>(_ ff: Kind<ProgramPartial<F>, (A) -> B>, _ fa: Kind<ProgramPartial<F>, A>) -> Kind<ProgramPartial<F>, B> {
+    public static func ap<A, B>(_ ff: ProgramOf<F, (A) -> B>, _ fa: ProgramOf<F, A>) -> ProgramOf<F, B> {
         Program(asFree: ff^.asFree.ap(fa^.asFree)^)
     }
 }
 
 extension ProgramPartial: Monad {
-    public static func flatMap<A, B>(_ fa: Kind<ProgramPartial<F>, A>, _ f: @escaping (A) -> Kind<ProgramPartial<F>, B>) -> Kind<ProgramPartial<F>, B> {
+    public static func flatMap<A, B>(_ fa: ProgramOf<F, A>, _ f: @escaping (A) -> ProgramOf<F, B>) -> ProgramOf<F, B> {
         Program(asFree: fa^.asFree.flatMap { f($0)^.asFree }^)
     }
 
-    public static func tailRecM<A, B>(_ a: A, _ f: @escaping (A) -> Kind<ProgramPartial<F>, Either<A, B>>) -> Kind<ProgramPartial<F>, B> {
+    public static func tailRecM<A, B>(_ a: A, _ f: @escaping (A) -> ProgramOf<F, Either<A, B>>) -> ProgramOf<F, B> {
         Program(asFree: Free.tailRecM(a, { f($0)^.asFree })^)
     }
 }
--- a/Sources/BowFree/Program.swift
+++ b/Sources/BowFree/Program.swift
@@ -15,9 +15,11 @@ public typealias ProgramOf<F, A> = Kind<ProgramPartial<F>, A>
 /// As opposed to `Free`, `Program` does not require `F` to be a functor, which means that evaluation of `map` calls are deferred
 /// and left for the later interpretation in another monad.
 public final class Program<F, A>: ProgramOf<F, A> {
-    let asFree: Free<CoyonedaPartial<F>, A>
+    /// Internal representation of `Program`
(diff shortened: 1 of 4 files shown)


Sources/BowFree/Program.swift
```diff
@@ -0,0 +1,92 @@
...
+    /// Folds this program structure using the same Monad.
+    ///
+    /// - Returns: Folded value.
+    func run() -> Kind<F, A> {
+        self.foldMapK(FunctionK<F, F>.id)
+    }
+}
+
+extension ProgramPartial: Functor {
+    public static func map<A, B>(_ fa: Kind<ProgramPartial<F>, A>, _ f: @escaping (A) -> B) -> Kind<ProgramPartial<F>, B> {
```
truizlop, line 69:
```suggestion
    public static func map<A, B>(_ fa: ProgramOf<F, A>, _ f: @escaping (A) -> B) -> ProgramOf<F, B> {
```

Sources/BowFree/Program.swift
```diff
@@ -0,0 +1,92 @@
...
+        Program(asFree: fa^.asFree.map(f)^)
+    }
+}
+
+extension ProgramPartial: Applicative {
+    public static func pure<A>(_ a: A) -> Kind<ProgramPartial<F>, A> {
```
truizlop, line 75:
```suggestion
    public static func pure<A>(_ a: A) -> ProgramOf<F, A> {
```

Sources/BowFree/Program.swift
```diff
@@ -0,0 +1,92 @@
...
+        Program(asFree: Free<CoyonedaPartial<F>, A>.pure(a)^)
+    }
+
+    public static func ap<A, B>(_ ff: Kind<ProgramPartial<F>, (A) -> B>, _ fa: Kind<ProgramPartial<F>, A>) -> Kind<ProgramPartial<F>, B> {
```
truizlop, line 79:
```suggestion
    public static func ap<A, B>(_ ff: ProgramOf<F, (A) -> B>, _ fa: ProgramOf<F, A>) -> ProgramOf<F, B> {
```

Sources/BowFree/Program.swift
```diff
@@ -0,0 +1,92 @@
...
+        Program(asFree: ff^.asFree.ap(fa^.asFree)^)
+    }
+}
+
+extension ProgramPartial: Monad {
+    public static func flatMap<A, B>(_ fa: Kind<ProgramPartial<F>, A>, _ f: @escaping (A) -> Kind<ProgramPartial<F>, B>) -> Kind<ProgramPartial<F>, B> {
```
truizlop, line 85:
```suggestion
    public static func flatMap<A, B>(_ fa: ProgramOf<F, A>, _ f: @escaping (A) -> ProgramOf<F, B>) -> ProgramOf<F, B> {
```

Sources/BowFree/Program.swift
```diff
@@ -0,0 +1,92 @@
...
+        Program(asFree: fa^.asFree.flatMap { f($0)^.asFree }^)
+    }
+
+    public static func tailRecM<A, B>(_ a: A, _ f: @escaping (A) -> Kind<ProgramPartial<F>, Either<A, B>>) -> Kind<ProgramPartial<F>, B> {
```
truizlop, line 89:
```suggestion
    public static func tailRecM<A, B>(_ a: A, _ f: @escaping (A) -> ProgramOf<F, Either<A, B>>) -> ProgramOf<F, B> {
```

Sources/BowFree/Program.swift
```diff
@@ -0,0 +1,92 @@
...
+
+/// Higher Kinded Type alias to improve readability.
+public typealias ProgramOf<F, A> = Kind<ProgramPartial<F>, A>
+
+/// Program is a type that, given any type constructor, is able to provide a Monad instance, that can be interpreted into a more restrictive one.
+///
+/// As opposed to `Free`, `Program` does not require `F` to be a functor, which means that evaluation of `map` calls are deferred
+/// and left for the later interpretation in another monad.
+public final class Program<F, A>: ProgramOf<F, A> {
+    let asFree: Free<CoyonedaPartial<F>, A>
```
truizlop, line 18: Would it make sense to expose this as a public field? I'm thinking about a case where we want to implement an instance of a type class for Program and we need to access this.

truizlop (approved):
Just some minor comments. Overall this is excellent, and at the same time almost trivial thanks to the work in the previous PRs!

ferranpujolcamins:
Thanks for the review. I keep forgetting about using *Of aliases!

Sources/BowFree/Program.swift
ferranpujolcamins: You are right!

ferranpujolcamins:
What do you think about the name?

truizlop (approved):

truizlop:
I think it is a good name, I like it better than Freer.

Issue #625, opened 2020-10-06
["Request"] Monad, Applicative and Functor instances for Trampoline


ferranpujolcamins:
## Description

Trampoline already has a `flatMap` function. By giving the class instances for Monad, Applicative and Functor we gain all the helper functions and combinators that come with this typeclasses.

I believe this has the potential to make `tailRecM` implementations easier to understand, by allowing us to build smaller Trampolines.
For example we will be able to cosntruct two or more Trampolines and combine them with Applicatives' map. 

## Modules

Bow

## Breaking changes

I believe none.

ferranpujolcamins:
Maybe also Traverse?

truizlop:
Sounds good to me. Actually, I've needed Traverse on Trampoline more than a couple of times. Are you taking this as well?

ferranpujolcamins:
Yes! :)

ferranpujolcamins:
@truizlop 

In order to add any typeclass instance to `Trampoline` we first need to add support for the higher kinded types simulation. Since this forces `Trampoline` to be a final class, the current *enum case as a subclass* approach won't work. We need to change to a *wrapped enum* approach as done in `Free`.

For the cases `Done` and `Defer`, the change is straightforward. For `FlatMap` no so much, because the subclassing mechanism was used to hide its type parameter `A`.

My proposed solution is to use `Coyoneda`. If you have closer look you'll see that `FlatMap` has the same data than `Coyoneda<ForTrampoline, Trampoline<B>>`, but `Coyoneda` hides the `A` parameter for us.

[Here](https://github.com/ferranpujolcamins/bow/commit/154cb006ae09ab7c7644ccee00e323f23639dfb8) ([trampolineTypeclasses](https://github.com/ferranpujolcamins/bow/tree/trampolineTypeclasses) branch) I implemented this approach successfully. However, the branch is based on #623 since I prefer the way `Coyoneda` is handled there (with polymorphic functions instead of casts). If that PR is finally not approved in a short time, I will of course rewrite the branch to work with the current `Coyoneda` implementation, I'm not trying to put pressure on you to approve the other PR. 

Until we decide on which approach to follow, I'll keep this on hold.

truizlop:
I think this is now ready to continue!

truizlop:
Closed by #636

PR #636, opened 2020-10-19
Trampoline typeclasses

ferranpujolcamins:
## Breaking changes
The result of `map` and `flatMap` now needs to be fixed with `^`. 

## Related issues

#625


Sources/Bow/Data/Trampoline.swift
```diff
@@ -40,78 +54,136 @@ public class Trampoline<A> {
...
+        }
     }
 }
 
-private final class Defer<A>: Trampoline<A> {
-    let deferred: () -> Trampoline<A>
+private final class FlatMapStep<A>: CokleisliK<CoyonedaFPartial<ForTrampoline, Trampoline<A>>, Either<() -> Trampoline<A>, A>> {
+    override init() {}
+
+    override func invoke<X>(_ fa: Kind<CoyonedaFPartial<ForTrampoline, Trampoline<A>>, X>) -> Either<() -> Trampoline<A>, A> {
```
truizlop, line 118:
```suggestion
    override func invoke<X>(_ fa: CoyonedaFOf<ForTrampoline, Trampoline<A>, X>) -> Either<() -> Trampoline<A>, A> {
```

truizlop (approved):
Everything looks good to me. I see you have moved Coyoneda to the main module; I am wondering if we should move Yoneda too, to maintain consistency, and keep only Free and Cofree in BowFree. What do you think?

ferranpujolcamins:
> Everything looks good to me. I see you have moved Coyoneda to the main module; I am wondering if we should move Yoneda too, to maintain consistency, and keep only Free and Cofree in BowFree. What do you think?

Sounds good to me. Let me do that in another PR.

ferranpujolcamins:
@truizlop can this be merged?

PR #642, opened 2020-10-21
Introduce HashableK typeclass

ferranpujolcamins:
## Goal

Let values of appropriate higher-kinded types be hashable.

I want this because in order to generate an `ArrowOf<KindOf<F, A>, B>` in a property-based test, `KindOf<F, A> ~ Kind<F, A>` needs to be hashable. With this we'll be able write property-based tests for `ComonadTrans` (PR coming soon).

### DIFF
--- a/Sources/Bow/Data/Try.swift
+++ b/Sources/Bow/Data/Try.swift
@@ -22,14 +22,14 @@ public typealias TryOf<A> = Kind<ForTry, A>
 
 /// Describes the result of an operation that may have thrown errors or succeeded. The type parameter corresponds to the result type of the operation.
 public final class Try<A>: TryOf<A> {
-    private let value: _Try<A>
+    fileprivate let value: Either<Error, A>
 
     /// Creates a successful Try value.
     ///
     /// - Parameter value: Value to be wrapped in a successful Try.
     /// - Returns: A `Try` value wrapping the successful value.
     public static func success(_ value: A) -> Try<A> {
-        Try(.success(value))
+        Try(.right(value))
     }
     
     /// Creates a failed Try value.
@@ -37,7 +37,7 @@ public final class Try<A>: TryOf<A> {
     /// - Parameter error: An error.
     /// - Returns: A `Try` value wrapping the error.
     public static func failure(_ error: Error) -> Try<A> {
-        Try(.failure(error))
+        Try(.left(error))
     }
 
     /// Creates a failed Try value.
@@ -61,7 +61,7 @@ public final class Try<A>: TryOf<A> {
         }
     }
 
-    private init(_ value: _Try<A>) {
+    private init(_ value: Either<Error, A>) {
         self.value = value
     }
 
@@ -80,15 +80,13 @@ public final class Try<A>: TryOf<A> {
     ///   - fa: Closure to apply if the contained value in this `Try` is a successful value.
     /// - Returns: Result of applying the corresponding closure to this value.
     public func fold<B>(_ fe: (Error) -> B, _ fa: (A) throws -> B) -> B {
-        switch value {
-        case let .success(a):
+        value.fold(fe, { a in
             do {
                 return try fa(a)
             } catch let error {
                 return fe(error)
             }
-        case let .failure(e): return fe(e)
-        }
+        })
     }
 
     /// Checks if this value is a failure.
@@ -189,11 +187,6 @@ public postfix func ^<A>(_ fa: TryOf<A>) -> Try<A> {
     Try.fix(fa)
 }
 
-private enum _Try<A> {
-    case success(A)
-    case failure(Error)
-}
-
 // MARK: Conformance of Try to CustomStringConvertible
 extension Try: CustomStringConvertible where A: CustomStringConvertible {
     public var description : String {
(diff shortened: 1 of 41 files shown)


Tests/BowLaws/HashableKLaws.swift
```diff
@@ -0,0 +1,22 @@
...
+    private static func coherenceWithEquality() {
+        property("Equal objects have equal hash") <~ forAll { (fa1: KindOf<F, A>, fa2: KindOf<F, A>) in
+            return (fa1.value == fa2.value) ==> {
+                var hasher1 = Hasher()
+                var hasher2 = Hasher()
+                fa1.value.hash(into: &hasher1)
+                fa2.value.hash(into: &hasher2)
+                return hasher1.finalize() == hasher2.finalize()
+            }
+        }
```
truizlop, line 20: I think this property will rarely find a mistake in the implementation, as I guess the chance of generating two equal inputs `fa1` and `fa2` is low, therefore the right hand side of the implication will rarely be invoked.

Sources/Bow/Data/Validated.swift
```diff
@@ -14,9 +14,9 @@ public typealias ValidatedNEA<E, A> = Validated<NEA<E>, A>
 
 /// Validated is a data type to represent valid and invalid values. It is similar to `Either`, but with error accumulation in the invalid case.
 public final class Validated<E, A>: ValidatedOf<E, A> {
-    private let value: _Validated<E, A>
+    fileprivate let value: Either<E, A>
```
truizlop, line 17: Although it removes some sort of duplication and it is just used internally, I think it is convenient to maintain the `valid`/`invalid` cases in Validated.

Sources/Bow/Data/Try.swift
```diff
@@ -223,6 +216,14 @@ extension TryPartial: EquatableK {
     }
 }
 
+// MARK: Instance of HashableK for Try
+extension TryPartial: HashableK {
+    public static func hash<A>(_ fa: TryOf<A>, into hasher: inout Hasher) where A : Hashable {
+        fa^.value.foldRun { hasher.combine("\($0)") }
+                       _: { hasher.combine($0) }
```
truizlop, line 223: Please, use the parenthesis versions to try to maintain compatibility with previous Swift versions.

Sources/Bow/Data/Try.swift
```diff
@@ -22,22 +22,22 @@ public typealias TryOf<A> = Kind<ForTry, A>
 
 /// Describes the result of an operation that may have thrown errors or succeeded. The type parameter corresponds to the result type of the operation.
 public final class Try<A>: TryOf<A> {
-    private let value: _Try<A>
+    fileprivate let value: Either<Error, A>
```
truizlop, line 25: Like in the case of `Validated`, I prefer maintaining its internal representation.

truizlop (changes requested):

Tests/BowLaws/HashableKLaws.swift
ferranpujolcamins: Equal instances are generated ~350 times for the whole Bow test suite, so I guess it's testing something. Might not be the most efficient test, but I think It does no harm to keep it.

Sources/Bow/Data/Validated.swift
ferranpujolcamins: I understand, will revert.

ferranpujolcamins:
Issues addressed.

Tests/BowLaws/HashableKLaws.swift
truizlop: I think we can use a custom generator to guarantee equal instances are generated in the same proportion as different instances in this particular property. We can make such custom generator as:

```swift
let generator: Gen<(KindOf<F, A>, KindOf<F, A>)> = {
        let equal = KindOf<F, A>.arbitrary.map { fa in (fa, fa) }
        let distinct = Gen.zip(KindOf<F, A>.arbitrary, KindOf<F, A>.arbitrary)
        return Gen.one(of: [equal, distinct])
    }()
```

Then, we can rewrite the property using:

```swift
property("Equal objects have equal hash") <~ forAllNoShrink(self.generator) { (tuple: (KindOf<F, A>, KindOf<F, A>)) in
```

Notice that I am passing the generator and using `forAllNoShrink`, as the tuple does not conform to `Arbitrary`. With this, we are forcing to have 50% chance of getting equal values for this property and strengthening its value.

(It may not be this exact syntax, I haven't tested this code on Xcode).

ferranpujolcamins: The problem I see with this approach is that we don't test pairs of objects that are different (as in different memory layout) but equal according to Equatable.

For example, if I had this type
```swift
struct DummyPair<A>: Equatable & Hashable {
    let first: A
    let second: A

    static func == (lhs: DummyPair, rhs: DummyPair) -> Bool {
        lhs.first == rhs.first
    }
}
```
the generator you propose would never generate the following case
```swift
DummyPair(1, 2) == DummyPair(1, 3)
// true
DummyPair(1, 2).hashValue == DummyPair(1, 3)
// false, because the default Hashable implementation will take into account the second value
```
and we wouldn't detect that we are missing a custom implementation of Hashable here.

truizlop: I see, what you mean, but isn't that covered in the `distinct` generator? That is the same generator we are using when none is passed explicitly. If you want, we can tweak the probabilities of using each generator (`equal` vs `distinct`) to be 0.3-0.7 instead of 0.5-0.5 or something like that.

ferranpujolcamins: Ah sorry, you are right, I read too fast. Looks good to me!

truizlop (approved):

truizlop:
@ferranpujolcamins There are conflicts on the project, can you address them?

ferranpujolcamins:
Done

truizlop:
Thanks!

PR #645, opened 2020-10-27
Monad trans

ferranpujolcamins:
## Goal

Be able to test Monad and Comonad transformers implementation.

## Implementation details

I added MonadTrans typeclass with its laws. I added ComonadTrans typeclass with its laws. Instances for each transformer are also included.


truizlop (approved):
Everything looks very good to me!

PR #649, opened 2020-11-02
Typed monad comprehensions

ferranpujolcamins:
## Goal

When using monad comprehensions, `flatMap` is called with values of type `Kind<F, Any>` and erased functions. From a result point of view, this is fine, as checked by the "Monad comprehensions equivalence to flatMap"  law.

However, when debugging, the situation is a little weird because if I set a breakpoint on the implementation of `flatMap` for a Monad, I cannot  know what's the inner type of my Monad at that point, because I always see `Kind<F, Any>`.

This PR fixes this.

## Implementation details

Instead of erasing to `Any`, `BindingExpression` now uses `Exists`. This allows us to call `flatMap` with the right type, instead of Any.

## Testing details

The test is a little bit weird, but I couldn't think of any other way to test this.


ferranpujolcamins:
I just realised that some unrelated changes slipped into this PR. I removed the unrelated commits and force-pushed.

truizlop (approved):
This will be very useful!

PR #650, opened 2020-11-05
Make Exists conform to CustomStringConvertible

ferranpujolcamins:
## Goal

Make `Exists` conform to `CustomStringConvertible` and `CustomDebugStringConvertible`.

## Implementation details

`Exists` might not always be truly `CustomStringConvertible`, because the inner type might not conform to `CustomStringConvertible`.

When `Exists` holds a value of an inner type that is not `CustomStringConvertible` it will use `\(fa)` as description.

For `CustomDebugStringConvertible` the behavior is the same.

### DIFF
--- a/Tests/BowTests/Data/ExistsTests.swift
+++ b/Tests/BowTests/Data/ExistsTests.swift
@@ -124,6 +124,25 @@ class ExistsTests: XCTestCase {
         }
     }
 
+    func testCustomDebugStringConvertibleWhenInnerTypeIs() {
+        property("""
+            Description matches inner type's description
+            when the inner type conforms to CustomStringConvertible
+        """) <~ forAll { (array: ArrayK<String>) in
+            return Exists(array).debugDescription == array.debugDescription
+        }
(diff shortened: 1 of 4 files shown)


Tests/BowTests/Data/ExistsTests.swift
```diff
@@ -98,6 +98,51 @@ class ExistsTests: XCTestCase {
...
+        CustomStringConvertibleLaws<Exists<ForArrayK>>.check()
+        CustomStringConvertibleLaws<Exists<ForId>>.check()
+        CustomStringConvertibleLaws<Exists<ForOption>>.check()
+    }
+
+    func testDescriptionMatchesInnerTypeDescription() {
+        property("""
+            Description matches inner type's description
+            when the inner type conforms to CustomStringConvertible
+        """) <~ forAll { (array: ArrayK<Int>) in
```
truizlop, line 111: Nice way of describing properties!

truizlop (approved):

PR #653, opened 2020-11-18
Remove Kinds from optics

truizlop:
## Goal

Remove Kinds from optics.

## Implementation details

The only necessary change was to remove the boilerplate associated to the declaration of the optics. This was never used and adds an overhead on the optics. As a result, some of the optics (except for Fold and Traversal) can be implemented as structs.


miguelangel-dev (approved):
LGTM

